PR #3, opened 2015-12-29
New load demo file -- draws inliers and outliers from two different s…

deepakn94:
…ets of normal distributions.

Let me know if what you had something else in mind! :)


pbailis:
LGTM! We can configure later when we're benchmarking.

PR #4, opened 2015-12-31
Preliminary timing

deepakn94:
Not quite done yet :)


pbailis:
LGTM to merge now if you want to.

It would be good to add more timing information to FPGrowthStreaming, including restructuring time. That code is particularly rough and probably deserves some discussion.

deepakn94:
@pbailis: Added some more timing. Going to merge this in. As we run experiments, we can more as needed.

PR #5, opened 2016-01-01
Implementation of data cache for the server and GUI.

pbailis:
Speeds up loading time by re-using previously loaded datasets. Not great quality, but does its job.


pbailis:
Also fixes bug in overflowing HTML tables.

PR #6, opened 2016-01-08
Workflow generation

deepakn94:
Creates workflow according to provided parameters, then runs workflow and parses out some basic runtime statistics about how long the workflow takes in different stages.
Next step involves parsing out number of itemsets detected.


deepakn94:
Ok, I'll switch timing to microseconds to get better accuracy. I thought about the timing overhead -- we can try to do something smarter in a next iteration, if it proves to be a problem.
And I'll take a stab at the inlier and outlier counts. :)

Issue #9, opened 2016-01-11
Re-scale metrics upon ingest.


pbailis:
For consistency across datasets and possibly improved convergence time, re-scale our metrics to a [0, 1] scale upon ingest. To start, let's do (x-min(X))/(max(X)-min(X)).

pbailis:
We may want to revisit this if it doesn't fix some of the weird C-Step behavior in MCD we were seeing.

deepakn94:
C-Step behavior?

pbailis:
The inner loop in FastMCD: https://github.com/stanford-futuredata/macrobase/blob/master/src/main/java/macrobase/analysis/outlier/MinCovDet.java#L209

PR #19, opened 2016-01-12
Rescale Datum metrics to [0, 1]. Addresses #9.

pbailis:
@deepakn94 want to peek quickly before we merge?

### DIFF
--- a/src/main/java/macrobase/ingest/SQLLoader.java
+++ b/src/main/java/macrobase/ingest/SQLLoader.java
@@ -129,6 +129,9 @@ public abstract class SQLLoader {
 
         List<Datum> ret = Lists.newArrayList();
 
+        RealVector metricWiseMinVec = new ArrayRealVector(lowMetrics.size()+highMetrics.size());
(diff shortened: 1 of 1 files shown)


src/main/java/macrobase/ingest/SQLLoader.java
```diff
@@ -141,17 +144,57 @@ public RowSet getRows(String baseQuery,
...
+                }
+
+                if(metricWiseMaxVec.getEntry(vecPos) < val) {
+                    metricWiseMaxVec.setEntry(vecPos, val);
+                }
+
+                vecPos += 1;
             }
 
             for(; i <= attributes.size()+lowMetrics.size()+highMetrics.size(); ++i) {
```
deepakn94: Hm, I think you want to set `i = attributes.size()+1` here

pbailis: Where, exactly? Sorry, confused.

deepakn94: Actually, scratch that.
Looks good.

PR #36, opened 2016-01-20
Make detector class configurable.

pbailis:
Defaults to MAD for 1 dimension, MCD for 2+. Preserves backwards compatibility. @deepakn94

### DIFF
--- a/src/main/java/macrobase/analysis/BaseAnalyzer.java
+++ b/src/main/java/macrobase/analysis/BaseAnalyzer.java
@@ -1,16 +1,30 @@
 package macrobase.analysis;
 
+import macrobase.analysis.outlier.MAD;
+import macrobase.analysis.outlier.MinCovDet;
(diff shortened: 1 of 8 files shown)


src/main/java/macrobase/analysis/BaseAnalyzer.java
```diff
@@ -43,4 +57,29 @@ public void setAlphaMCD(double alphaMCD) {
     public void setStoppingDeltaMCD(double stoppingDeltaMCD) {
     	this.stoppingDeltaMCD = stoppingDeltaMCD;
     }
+
+    protected OutlierDetector constructDetector(int metricsDimensions) {
+        if(detectorType == null) {
```
deepakn94: How else might detectorType be set?

deepakn94: From the configuration file?

PR #43, opened 2016-01-21
Allow users to specify dbUser and dbPassword in config

mamikonyana:


.gitignore
```diff
@@ -35,3 +36,6 @@ bench/workflows/*
 
 # Annoying Mac file
 *.DS_Store
+
+# Arsen's local server configuration
+conf/arsen*
```
pbailis: Hmm, not sure we want this in the main source tree.

src/main/java/macrobase/runtime/MacroBaseServer.java
```diff
@@ -46,6 +46,7 @@ public void run(ServerConfiguration configuration,
 
         if(configuration.doPreLoadData() && (!configuration.getPreLoad().isEmpty() || true)) {
             for(ServerConfiguration.PreLoadData pld : configuration.getPreLoad()) {
+                loader.setDatabaseCredentials(pld.getDbUser(), pld.getDbPassword());
```
pbailis: Does this still work if dbUser and dbPassword are not set?

.gitignore
mamikonyana: where else? I just put this in the gitignore to have a clean working copy when changing my configuration

src/main/java/macrobase/runtime/MacroBaseServer.java
mamikonyana: yes it does

.gitignore
pbailis: What about just something like `conf/custom`? As it stands, everyone who downloads the code will have a special case for `conf/arsen`. :)

mamikonyana:
I updated and now there is `conf/custom` in the .gitigrnoe

pbailis:
Great, thanks! Feel free to merge via @deepakn94's workflow [on the wiki](https://github.com/stanford-futuredata/macrobase/wiki/Development-Notes#git-workflow) (pushing will close the PR).

mamikonyana:
I just pushed to master, but nothing happened to this pull request. Should I force push to this branch too? That might also delete all the comments on the code.

deepakn94:
Nope it's fine, no need to push to this branch, I'm just going to manually close this PR. (the reason it didn't close was because this local branch and the master branch diverged -- that is, the new master branch is not simply the old master branch + this branch on top of that)

mamikonyana:
Right, but that would happen any time there is a rebase, no?

deepakn94:
Right. I guess if you wanted to this the "right way", you would do what you alluded to earlier where you first force push to the remote branch, and then merge --ff-only on your local master branch before pushing.

PR #46, opened 2016-01-24
Add option to disable tracing in inner loop for streaming.

pbailis:
This PR makes fine-grained timing for the streaming inner loop configurable (default: off). Since Java doesn't have pre-processor macros, there were a few ways to go. This is one of the simplest but the least generic.

@deepakn94: any thoughts?

### DIFF
--- a/src/main/java/macrobase/analysis/StreamingAnalyzer.java
+++ b/src/main/java/macrobase/analysis/StreamingAnalyzer.java
@@ -81,6 +81,10 @@ public class StreamingAnalyzer extends BaseAnalyzer {
         this.minRatio = minRatio;
     }
 
+    public void setTracing(boolean doTrace) {
+        this.doTrace = doTrace;
+    }
+
     public void setOutlierItemSummarySize(Integer outlierItemSummarySize) {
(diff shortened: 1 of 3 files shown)


deepakn94:
The optimization works even though `traceRuntime` is set at runtime? [through the configuration file]

pbailis:
Yep, it's kind of a hack, but check out: https://github.com/stanford-futuredata/macrobase/blob/f76a465530a4bcac6cc525908a1669da9a476168/src/main/java/macrobase/analysis/StreamingAnalyzer.java#L199

deepakn94:
I see, I was just unsure whether the Java compiler was smart enough to optimize that out

pbailis:
No, it won't be optimized out but it should be cheap. An alternative is to have an alternative, static member as in `static boolean do_trace = false` and corresponding set of methods like `startTimer(Timer t) { if(do_trace) { t.start() }` for each of starting, ending, resetting the timers.

deepakn94:
Right, this seems like a tradeoff: harder to set the static member, for better efficiency.

pbailis:
Are you okay with committing this?

deepakn94:
Yup, sure. LGTM

PR #47, opened 2016-01-24
Connect to istc9 database with custom user

mamikonyana:
It seems that I have to be `pbailis` to be able to connect to the database, postgres user didn't work


pbailis:
LGTM, though the commit message isn't very descriptive. Perhaps "Enable login using custom username and password."

PR #50, opened 2016-01-26
Don't want to be using KDE by default yet :)

deepakn94:
@pbailis @mamikonyana


pbailis:
LGTM, thanks!

@mamikonyana -- until this outperforms MCD and MAD, we'll want to keep it off the fast path.

Issue #51, opened 2016-01-26
Batch scoring by percentile is swapped.


pbailis:
https://github.com/stanford-futuredata/macrobase/blob/master/src/main/java/macrobase/analysis/outlier/OutlierDetector.java#L88

deepakn94:
Handled by #52

PR #52, opened 2016-01-26
Fix to outlierDetector: make sure to pick split point from the right …

deepakn94:
…side
@pbailis 

Addresses #51


pbailis:
LGTM, thanks!

PR #56, opened 2016-01-26
Add a way to specify password/user for benchmarking scripts…

mamikonyana:
…also tweaks to parameter storage

@deepakn94 could you take a look?


mamikonyana:
I moved configurations into their own `.json` files, i hope you don't mind

deepakn94:
What's the behavior like if the `USER` environment variable isn't set?
Otherwise, LGTM.

mamikonyana:
I didn't test that, I'll add a fail safe to default to postgres. but `USER` is set automatically on linux and Mac, so that shouldn't be a problem.

mamikonyana:
I actually set the default to None, in that case dbUser will be ignored in the config file.

deepakn94:
Alright, sounds good. Thanks!

PR #59, opened 2016-01-27
Tweaks to be able to compare different type analyzers

mamikonyana:


bench/execute_workflows.py
```diff
@@ -92,13 +91,15 @@ def parse_results(results_file):
...
+
 def get_stats(value_list):
   value_list = [float(value) for value in value_list]
   mean = sum(value_list) / len(value_list)
   stddev = (sum([(value - mean)**2 for value in value_list]) / len(value_list)) ** 0.5
   return mean, stddev
 
-def run_workload(config_parameters, print_itemsets=True):
+
+def run_workload(config_parameters, repeats, print_itemsets=True):
```
deepakn94: Let's rename this to `num_runs`.

bench/execute_workflows.py
```diff
@@ -152,8 +153,7 @@ def run_workload(config_parameters, print_itemsets=True):
   print "Mean tuples / second:", mean_tuples_per_second, ", Stddev tuples / second:", stddev_tuples_per_second
 
 
-def run_all_workloads(configurations,
-                      script_arguments={},
+def run_all_workloads(configurations, defaults, repeats_per_workload,
```
deepakn94: Here too.

bench/execute_workflows.py
```diff
@@ -198,6 +193,11 @@ def parse_args():
                       type=argparse.FileType('r'),
                       default='conf/sweeping_parameters.json',
                       help='File with a dictionary of sweeping parameters')
+  parser.add_argument('--workload-repeats', default=5, type=int,
```
deepakn94: `num_runs`

deepakn94:
Sorry for the nitpicks!

mamikonyana:
it's good. I even like that!

deepakn94:
Actually this leads to undesired behavior. We currently _want_ to not explicitly specify an `outlierDetectorType` when we run experiments, since we want to dynamically pick an outlier detection algorithm depending on the dimension of the input points.

deepakn94:
Better behavior would be to plug in the `outlierDetectorType` into the configuration file only if `outlierDetectorType` is provided in the configuration dict.

pbailis:
We could make specifying the `outlierDetectorType` required. I think eventually we want to. However, our scripts don't currently support it.

deepakn94:
Yeah, I guess. It's a pain to go ahead and pick between MAD and MCD for every workload. (especially when that information can be inferred directly from the dimensionality of the input points)

PR #64, opened 2016-02-03
Timeseries - Moving Average Detector

viveksjain:
As promised, time series detector along with code to run the Numenta benchmark. And the results are in for moving average! *drumroll*
Final score for 'macrobase' detector on 'ma_reward_low_FP_rate' profile = 0.68
Final score for 'macrobase' detector on 'ma_reward_low_FN_rate' profile = 15.22
Final score for 'macrobase' detector on 'ma_standard' profile = 5.59

Compared to [random](https://github.com/numenta/NAB#scoreboard) Reward Low FP = 5.8, Reward Low FN = 25.9, Standard Profile = 16.8. :)

### DIFF
--- a/script/numenta_benchmark.py
+++ b/script/numenta_benchmark.py
@@ -1,0 +1,137 @@
+import psycopg2
+import sys
+import os
+import os.path
+import csv
+import dateutil.parser
+from datetime import datetime
+import itertools
+import re
+import json
+
+# Note: initially need to run `python scripts/create_new_detector.py --detector macrobase_ma`
+# This script also requires additional logging statements in MacroBase - you need to use the nab branch.
+# Then run this script with an argument of the path to the numenta anomaly benchmark directory.
+
+# TODO we should probably tune window size
+# TODO should add config parameter to disable scoring, change training percentage (0.15 for NAB)
+config = '''
+macrobase.query.name: numenta
+macrobase.loader.attributes: [value]
+macrobase.loader.targetLowMetrics: []
+macrobase.loader.targetHighMetrics: [value]
+macrobase.loader.timeColumn: timestamp
+
+macrobase.loader.db.dbUrl: postgres
+macrobase.loader.db.baseQuery: SELECT * FROM nab;
+macrobase.loader.transformType: IDENTITY
+
+macrobase.analysis.detectorType: MOVING_AVERAGE
+
+macrobase.analysis.minSupport: 0.001
+macrobase.analysis.minOIRatio: 1
+
+macrobase.analysis.useZScore: false
+
+macrobase.analysis.usePercentile: true
+macrobase.analysis.targetPercentile: 0.99
+
+macrobase.analysis.timeseries.tupleWindow: 100
+
+logging:
+  level: WARN
+
+  loggers:
+    macrobase.analysis.outlier: DEBUG
+'''
+conf_file = '/tmp/macrobase_nab.conf'
+result_file = '/tmp/macrobase_results.out'
+mb_detector_name = 'macrobase_ma'
+
+def main():
+    if len(sys.argv) != 2:
+        print 'Usage: numenta_benchmark.py PATH_TO_NAB_ROOT'
+        sys.exit(1)
+
+    with open(conf_file, 'w') as f:
+        f.write(config)
+
+    bench_dir = sys.argv[1]
+
+    with open(os.path.join(bench_dir, 'labels', 'combined_windows.json')) as jf:
+            label_windows = json.load(jf)
+            for path, labels in label_windows.iteritems():
+                label_windows[path] = map(lambda pair: map(tounixdate, pair), labels)
+
+    data_dir = os.path.join(bench_dir, 'data')
+    for dirpath, dirnames, filenames in os.walk(data_dir):
+        for f in filenames:
+            if not f.endswith('.csv'):
+                continue
+            reldir = os.path.relpath(dirpath, data_dir)
+            print reldir
+            path = os.path.join(dirpath, f)
+            relpath = os.path.join(reldir, f)
+            outpath = os.path.join(bench_dir, 'results', mb_detector_name, reldir, '%s_%s' % (mb_detector_name, f))
+            run(path, outpath, label_windows[relpath])
+
+    print 'Done! Inside the NAB directory run `python run.py -d %s --optimize --score --normalize`' % mb_detector_name
+
+def run(csvpath, outpath, labels):
+    conn = psycopg2.connect("dbname='postgres' host='localhost'")
+    cur = conn.cursor()
+    cur.execute("DROP TABLE IF EXISTS nab; CREATE TABLE nab ( timestamp bigint NOT NULL, value numeric NOT NULL );")
+
+    print 'Loading ' + csvpath
+    with open(csvpath, 'rb') as in_csv:
+        reader = csv.reader(in_csv)
+        for row in skip_first(reader):  # Skip header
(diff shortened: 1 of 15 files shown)


viveksjain:
Also, there's a couple of TODOs still in the code - I'm going to try to fix up most of them in the next couple of weeks.

pbailis:
Planning to look this over tonight.

script/numenta_benchmark.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import os.path
+import csv
+import dateutil.parser
+from datetime import datetime
+import itertools
+import re
+import json
+
+# Note: initially need to run `python scripts/create_new_detector.py --detector macrobase_ma`
```
pbailis: What's the long-term gameplan for the `nab` branch?

src/main/java/macrobase/analysis/outlier/MovingAverage.java
```diff
@@ -0,0 +1,88 @@
...
+        windowSum = null;
+    }
+
+    @Override
+    public void trainWindow(Queue<Datum> data, Datum newDatum, Datum oldDatum) {
+        /* Empty */
+    }
+
+    @Override
+    public double scoreWindow(Queue<Datum> data, Datum newDatum, Datum oldDatum) {
```
pbailis: So, this is basically a tuple-based sliding window model, where `oldDatum` is what just exited the window. There's another way to do this so you talk about windows composed of `panes`, where each pane can contain multiple transactions and the window slides pane by pane.

src/main/java/macrobase/ingest/SQLLoader.java
```diff
@@ -88,6 +93,7 @@ public RowSet getRows(String baseQuery,
                           List<RowSetResource.RowSetRequest.RowRequestPair> preds,
                           int limit,
                           int offset) throws SQLException {
+        // TODO do we need to handle time column here?
```
pbailis: What are you worried about?

src/main/java/macrobase/ingest/SQLLoader.java
```diff
@@ -213,5 +233,19 @@ public RowSet getRows(String baseQuery,
 
         return ret;
     }
+    
+    // Shield your eyes, mere mortals, from this glorious hideousness.
```
pbailis: :100:

pbailis:
This is looking good. My main two comments are:

 1. Let me know if this "pane" idea doesn't make sense. It's marginally more general and will allow us to have arbitrarily-sized window slides.
 2. I'm not a huge fan of having timeColumn all over the code. It may be the cleanest way in the end, but, nevertheless, have you thought about alternatives?

I liked how you handled the `OutlierDetector` inheritance.

script/numenta_benchmark.py
viveksjain: That's a good question. slf4j would have been ideal since it's already being used, but unfortunately it's buggy. The options I see are: (i) keep the nab branch, updating as necessary. I don't think it's a huge deal if it gets out of date. (ii) Adding another flag. I'm not a fan of having a ton of different flags, but if we want to use this longer term it would probably be the better approach.

src/main/java/macrobase/analysis/outlier/MovingAverage.java
viveksjain: Makes sense, but seems like a slightly more complicated API. Is there a model that requires pane support? Otherwise, in my opinion, simplicity > unused generality.

src/main/java/macrobase/ingest/SQLLoader.java
viveksjain: Do we need to order by the time column here as well? Also, the approach in this file seems rather hacky in general - I'm currently assuming `baseQuery` doesn't already contain the `ORDER BY` etc etc. But I guess it's good enough for our purposes for now.

src/main/java/macrobase/ingest/SQLLoader.java
viveksjain: This is how I vent.

viveksjain:
Re: 2. I don't know what the alternatives are. If you have other ideas, I'm all ears. The only thing I can think of is that `timeColumn, attributes, lowMetrics, highMetrics` are used together in a bunch of places - maybe it's worth refactoring them out into a single wrapper class.

script/numenta_benchmark.py
pbailis: I'd love to have numenta working long-term. What's buggy about slf4j? I'm happy to migrate if you have a preferred alternative.

src/main/java/macrobase/analysis/outlier/MovingAverage.java
pbailis: The main model I want to support is real-time sliding window queries -- e.g., give me the EWMA over the last five minutes. If you think the tuple-at-a-time interface is a good fit for expressing this, I'm okay with it.

src/main/java/macrobase/ingest/SQLLoader.java
pbailis: Yeah, `baseQuery` is a hack. As an intermediate option, you could check for `ORDER BY` and error out.

src/main/java/macrobase/ingest/SQLLoader.java
pbailis: :+1:

src/main/java/macrobase/analysis/outlier/MovingAverage.java
viveksjain: There's a distinction to be made between the interface for algorithm writers and for algorithm users. For algorithm writers, I think it's generally easier to think in terms of tuple-at-a-time, and this interface is designed for them. For consumers of the algorithm's scoring, it may be easier to use the panes idea, but that can be a different interface/class that calls the TimeSeriesOutlierDetector interface inside.

viveksjain: For posterity: Peter and I discussed offline and I realized the issue with this `scoreWindow` is essentially that it assumes the window is of a fixed size, and every time you remove a tuple you also add another one. This may not be the case if the window, for example, represents a fixed time range and the points are not uniformly spread out. This can be fixed by having interfaces that allow individually adding/removing one (or many) tuples at a time.

viveksjain:
@pbailis good to merge for now? Main remaining TODOs: wrapper class for `timeColumn, attributes, lowMetrics, highMetrics`, tests, logging for numenta. For the last one, I can show you the log4j bug in person. Best approach is probably to add a flag; it would be nicer to do that after #76 so that I don't have to pass around the flag in a bunch of unnecessary places.

pbailis:
Can you add some basic tests? :100: 

Then :+1:

viveksjain:
Rebased and added basic tests. I decided to create DatumWithTime subclass - perhaps this is premature optimization, but this way we don't have to pay cost of the additional pointer in every Datum if the time isn't needed.

pbailis:
Great work! Can you squash for review?

viveksjain:
How come coveralls doesn't show `MovingAverage.java`?

pbailis:
Very strange, not sure...

Do you need me to take another look at this?

viveksjain:
Squashed!

pbailis:
Great. LMK if you need another look, otherwise feel free to merge.

viveksjain:
The server seems to still work (although of course, the web UI doesn't yet support time column) and the output matches output from master. So I'm going to be bold and

deepakn94:
Ur comment game is strong.

pbailis:
:+1: :100:

PR #66, opened 2016-02-05
Redirect stdout and stderr to results file.

pbailis:
@deepakn94 Does this break your analysis scripts?


deepakn94:
Shouldn't. Let's leave this open for now though. I'll test this and then merge if everything looks ok.

deepakn94:
Verified that this seems ok.

PR #68, opened 2016-02-07
Automatically set bandwidth (KDE), only specify the algorithm

mamikonyana:


mamikonyana:
Normal Scale picks a bandwidth that's too high, so it oversmoothes, but good bandwidth selectors are too computationally intensive.

PR #70, opened 2016-02-08
Scoping

raininthesun:
A first scoping implementation. Use need to specify 6 parameters as shown in scoping.yaml using uk_road_accidents dataset. 

I've also implemented the finding explanations for scoped outliers, which by default is turned off. That piece of code is in ScopingAnalyzer.java


conf/scoping.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+  loggers:
+    "macrobase": TRACE
+
+
+
+# The following configurations are for scoping
+categoricalAttributes : [police_force, day_of_week, local_authority_district,local_authority_highway, n_1st_road_class, road_type, junction_detail, junction_control, n_2nd_road_class, pedestrian_crossinghuman_control, pedestrian_crossingphysical_facilities, light_conditions, weather_conditions, road_surface_conditions, special_conditions_at_site, carriageway_hazards, urban_or_rural_area, did_police_officer_attend_scene_of_accident]
+numericalAttributes: [accident_severity, number_of_vehicles, speed_limit]
+numInterval: 9
```
pbailis: Interval is for discretizing the continuous metrics?

src/main/java/macrobase/runtime/standalone/scoping/SubSpace.java
```diff
@@ -0,0 +1,341 @@
...
+	
+	/**
+	 * The metric dimensions
+	 */
+	List<Integer> metricDimensions;
+	
+	
+	/**
+	 * The following fields are specific to an outlier detection method, right now it is count-based
+	 */
```
pbailis: Do you have a sense of how this will evolve over time?

src/main/java/macrobase/runtime/standalone/scoping/SubSpace.java
```diff
@@ -0,0 +1,341 @@
...
+	}
+	
+	/**
+	 * Return 
+	 * @param other
+	 * @param total
+	 * @param tau
+	 * @return
+	 */
+	public SubSpace join(SubSpace other,int total, double tau){
```
pbailis: What is tau? Can you add a comment?

pbailis:
Some issues to discuss on Thursday:

 - What's the long-term plan for scoping? I'd much prefer to have it as part of the main streaming and batch rather than as a separate command.
 - How is this code going to change when we consider non-count-based detection?
 - What's the plan for non-binned metrics?

conf/scoping.yaml
raininthesun: Interval is for discretizing numerical attributes, including scoping attributes, and metric attributes.  An extension would be dynamically figuring intervals for a numerical attribute.

src/main/java/macrobase/runtime/standalone/scoping/SubSpace.java
raininthesun: The following three fields are fixed for any outlier detection function f. 
List<Integer> dimensions;
List<Unit> denseUnits;
List<Integer> metricDimensions;


The following fields are specific to an outlier detection method, right now it is count-based, hence, the current structure of storing the information. 
Map<Unit,HashMap<Unit,List<Integer>>>  denseUnit2metricUnit2MergedTIDs = new HashMap<Unit,HashMap<Unit,List<Integer>>>();

src/main/java/macrobase/runtime/standalone/scoping/SubSpace.java
raininthesun: The minimum threshold for a Unit considered to be dense. Comment added in the code

pbailis:
Xu, LGTM, but can you rebase this into a single commit?

raininthesun:
I can do that for sure. But I think the current code of scoping is not the one we should merge into master, since it has a specific scoping implementation. 

I think I will work on a new branch that embeds the necessary data structure for contexts into the master data structure, and it should work for any type of outlier detection method. We will then merge that branch into master.  How does that sounds?

pbailis:
> But I think the current code of scoping is not the one we should merge into master, since it has a specific scoping implementation.

I agree completely. I wasn't sure if you wanted this merged for the time being. But if you're okay holding off, then yes, a generic/more general scoping implementation would be much better to merge.

PR #74, opened 2016-02-13
Allow specifying auxilaryAttributes 

mamikonyana:
which get dumped with the rest of data, very heplful for ploting performance

### DIFF
--- a/script/run_experiment.py
+++ b/script/run_experiment.py
@@ -14,7 +14,7 @@ from py_analysis.plot_estimator import plot_estimator
 
 def run_macrobase(cmd='batch', conf='conf/batch.conf', **kwargs):
   extra_args = ' '.join(['-Ddw.{key}={value}'.format(key=key, value=value)
-		         for key, value in kwargs.items()])
+                         for key, value in kwargs.items()])
   macrobase_cmd = '''java {extra_args} -Xms128m -Xmx16G \\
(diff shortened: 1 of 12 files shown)


pbailis:
Got it. Kind of messy, but probably okay. @deepakn94: thoughts?

@viveksjain is going to add another way to specify a column as time so I'm curious if there is a more generic way we want to handle this.

viveksjain:
From #64: `timeColumn, attributes, lowMetrics, highMetrics` are used together in a bunch of places - maybe it's worth refactoring them out into a single wrapper class.
It's not really a more generic way to handle this, more like duct tape to hide the excrescence. Maybe convert the rest to something like a DataTransformation so you have inversion of control? And then you could have have a wrapper class containing a List<DataTransformation>.

BTW, this is probably going to cause merge conflicts with #64, but since my branch already has merge conflicts I'm going to wait for this one to go in first.

viveksjain:
And I'd prefer any refactoring to be done after #64 is in.

mamikonyana:
Cool. I like the idea of putting all those variables in a wrapper.
I'll wait for a word from @deepakn94 and merge this after that.

deepakn94:
Sure, we can put these in a wrapper class.

pbailis:
Sure, we can do a refactoring push after #64.
@viveksjain: more incentive to get it in!

PR #75, opened 2016-02-16
Add BinnedKDE outlier detector; refactor Kernel into a class from an enum

mamikonyana:


pbailis:
Comments inline, plus two things:
 1. Consider squashing if the first commit doesn't stand alone.
 2. Consider a unit test or two!

mamikonyana:
First commit stands alone, I'll try to write a test

PR #82, opened 2016-02-22
Minor Changes

viveksjain:
Some small things I noticed while doing the timeseries rebase, and I figured it would be quicker and better style to put it in a separate branch.


pbailis:
I like the commit message sass.

LGTM; merge it!

viveksjain:
Also, I thought with the conf refactor we were going to pass around MacroBaseConf to each OutlierDetector so that it could construct itself however it wanted - this seems cleaner than BaseAnalyzer having to know about all of the params for every detector. @pbailis

pbailis:
Ah, this is a smart idea. Add as a new GH issue?

Issue #83, opened 2016-02-22
Pass MacroBaseConf to each OutlierDetector


viveksjain:
As discussed in #82. Holding off on implementing until #64 is in.

viveksjain:
See #84 for progress.

PR #84, opened 2016-02-22
Pass MacroBaseConf to each OutlierDetector

viveksjain:
Implement #83. BTW, it's also a little strange to me that MacroBaseConf needs to know how to construct DataLoaderType, DetectorType, BandwidthAlgorithm etc. But this is enough refactoring for today. :)

### DIFF
--- a/src/test/java/macrobase/analysis/outlier/MovingAverageTest.java
+++ b/src/test/java/macrobase/analysis/outlier/MovingAverageTest.java
@@ -15,7 +15,7 @@ import com.google.common.collect.ImmutableMap;
 public class MovingAverageTest {
     @Test
     public void simpleTest() {
-        MacroBaseConf conf = new MacroBaseConf(ImmutableMap.of(MacroBaseConf.TUPLE_WINDOW, 3));
+        MacroBaseConf conf = new MacroBaseConf().set(MacroBaseConf.TUPLE_WINDOW, 3);
         MovingAverage ma = new MovingAverage(conf);
         assertEquals(0, ma.score(createDatum(0, 1)), 0);
         assertEquals(1/3.0, ma.score(createDatum(1, 2)), 1e-5); // Average = 1.5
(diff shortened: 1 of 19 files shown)


viveksjain:
Gah I broke all the tests.

viveksjain:
mcdAlpha and mcdStoppingDelta from the conf didn't seem to be actually used earlier. Adding it in broke one of the tests (which I fixed) but I hope it doesn't break a previous workflow.

src/test/java/macrobase/analysis/outlier/MovingAverageTest.java
```diff
@@ -4,15 +4,19 @@
...
 import org.apache.commons.math3.linear.ArrayRealVector;
 import org.junit.Test;
 
+import com.google.common.collect.ImmutableMap;
+
 public class MovingAverageTest {
     @Test
     public void simpleTest() {
-        MovingAverage ma = new MovingAverage(3);
+        MacroBaseConf conf = new MacroBaseConf(ImmutableMap.of(MacroBaseConf.TUPLE_WINDOW, 3));
```
pbailis: Slightly prefer a `new MacroBaseConf().set(MacroBaseConf.TUPLE_WINDOW, 3)`. May require making `MacroBase.set` return `this`.

pbailis:
:100: Had one comment on style, that's unrelated.

viveksjain:
Agreed, I do like that style better.

pbailis:
An even fancier way is to add a static constructor:

`MacroBaseConf.create().set()...`

`public static MacroBaseConf create() { return new MacroBaseConf(); }`

viveksjain:
Hmm that's one of those "design patterns," isn't it *wrinkles nose in disgust*? Not that there's anything wrong with that, but I don't see the point in this case. If the static constructor was used to create a builder that is then immutable after being built, I understand why you would want to have it, but if it's just calling new then that's 2 ways of doing the same thing.

pbailis:
Fair -- let's merge this!

PR #86, opened 2016-02-22
Correct streaming prefix in bench.py

pbailis:
a0658a121a894fb05b613b846684d2fc361b226d had the wrong prefix.

### DIFF
--- a/bench/execute_workflows.py
+++ b/bench/execute_workflows.py
@@ -25,10 +25,10 @@ default_args = {
   "macrobase.analysis.streaming.inlierItemSummarySize": 10000,
   "macrobase.analysis.streaming.outlierItemSummarySize": 10000,
   "macrobase.analysis.streaming.summaryUpdatePeriod": 100000,
-  "macrobase.analysis.period.modelRefreshPeriod": 100000,
+  "macrobase.analysis.streaming.modelUpdatePeriod": 100000,
 
   "macrobase.analysis.streaming.useRealTimePeriod": "false",
-  "macrobase.analysis.period.useTupleCountPeriod": "true",
+  "macrobase.analysis.streaming.useTupleCountPeriod": "true",
 
   "macrobase.analysis.streaming.warmupCount": 50000,
   "macrobase.analysis.streaming.decayRate": 0.01,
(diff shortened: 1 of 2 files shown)


viveksjain:
LGTM! I like how coveralls fails if coverage stays the same...

viveksjain:
BTW, should we throw an error if an unrecognized flag is used so things like this can't happen?

deepakn94:
I like the idea of throwing an exception if we see an unrecognized flag!

pbailis:
So, this is untested because it's in our Python code.

My only reservation re: putting a check + exception in the Java is that it may limit extensibility. For example, if I add new functionality to a module (or implement my own third party OutlierDetector), now I'll need a separate conf file.

deepakn94:
I think I lost you there? My idea was to have a global list of correct configuration parameter names in the Java configuration class: if you add a new configuration parameter, then you would need to update the global list too.

viveksjain:
I think it's reasonable to ask for all flag names to be in `MacroBaseConf.java`, regardless of where it is used. However, I realize now that this could be annoying if you want to easily switch between versions where one version adds a new flag. At the very least, we could print a warning.

pbailis:
Yeah, I think a warning could be good...

OR

we could have a flag like `macrobase.conf.ignoreUnrecognizedParameters`.

Eh? @viveksjain @deepakn94

viveksjain:
A flag to disable flagging of unrecognized flags?

My general opinion, which you probably already know, is that there should be as few flags as possible - flags can interact in strange ways, and more flags means more code to maintain and more documentation for users to look through. And in this case, shouldn't the logging > loggers config option already handle this?

pbailis:
@viveksjain: cutting commentary... :100: 

If we want few flags and less hassle, isn't the simplest option to allow any flags?

viveksjain:
I probably have the least experience here in terms of working on large scale/long term projects, so take all of my opinion with a grain of salt :). But I would argue displaying warnings for unknown flags *is* less hassle _for the user_, so they know when they're doing the wrong thing.

deepakn94:
LGTM.

pbailis:
d63a8cb6b2e3b1c260e5a6b0c3c3645c706f07cb

PR #88, opened 2016-02-23
FPGrowth should traverse from bottom to top.

pbailis:
This should have been addressed in e2efa185f827412297b5cfb4429d626b701272d2

It was fixed in FPGrowthStreaming but not FPGrowth...
https://github.com/stanford-futuredata/macrobase/commit/e2efa185f827412297b5cfb4429d626b701272d2#diff-17a5f28a45018f3282473cce06fb484fR267

### DIFF
--- a/src/main/java/macrobase/analysis/summary/itemset/FPGrowth.java
+++ b/src/main/java/macrobase/analysis/summary/itemset/FPGrowth.java
@@ -33,7 +33,7 @@ public class FPGrowth {
         // used to calculate the order
         private Map<Integer, Double> frequentItemCounts = new HashMap<>();
 
-        // item order -- need canonical to break ties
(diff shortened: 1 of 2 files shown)


pbailis:
This is kind of a bizarro bug. I basically just need to test the FPGrowth code more and eliminate redundancy between the two.

The problem is that `frequentItemOrder` is sorted from smallest to largest. So we want to traverse the tree when mining in order, and insert into the tree in reverse order. Super ugly, and it turns out the two were inconsistent, yet somehow the test case worked for this...

The right thing to do is to add a new test case. I really want to refactor all of this. Maybe we can discuss consolidation of the streaming and batch pipelines during tomorrow's call.

deepakn94:
LGTM

pbailis:
427eb240b45dd35ae48595ff783528268068b195

PR #91, opened 2016-03-03
Dead code eliminination; Caching SQL loader test.

pbailis:
Should be an easy PR.

Eliminates dead code, adds a test to CachingSQLLoader.


mamikonyana:
I created a gzip file loader for my test, I'd recommend using that instead of checking in a 10K line long csv

pbailis:
@mamikonyana: fixed, thanks for the tip!

BTW, `src/test/resources/data/2d_standard_normal_100k.csv.gz` is 1.8MB compressed :) We may want to only use a sample...

pbailis:
Also, @mamikonyana is this good to go? :v:

mamikonyana:
lgtm

PR #93, opened 2016-03-04
ARIMA

viveksjain:
Implementation of ARIMA that calls out to R. I also have the dumb version of streaming MAD that recomputes on every window change, but given that we’re going to change the streaming API I don’t think that’s worth merging for now.

Note that there is also a [Spark ARIMA implementation](https://github.com/sryza/spark-timeseries/blob/master/src/main/scala/com/cloudera/sparkts/models/ARIMA.scala), which may be useful if we want a pure Java implementation, but it doesn't support many of the R version's features (e.g. seasonality, Kalman filtering, trying to automatically deduce p,d,q).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -180,5 +180,10 @@
       <version>1.1.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.lucarosellini.rJava</groupId>
+      <artifactId>JRI</artifactId>
+      <version>0.9-7</version>
+    </dependency>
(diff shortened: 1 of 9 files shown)


viveksjain:
BTW, I've added tests to keep the test gods happy, but they are skipped by default so that testing locally doesn't require R to work.

pbailis:
LGTM, but do you need to add a new configuration option?

viveksjain:
Hah, good catch. Merging once tests pass.

PR #96, opened 2016-03-06
New Design based on discussions on Thursday and Friday

mamikonyana:
I created a skeleton that shows a new API how components will fit into the pipeline, i.e. MacrobaseAnalyzer


src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private DataIngester ingester;
+    private FeatureTransformation transformation;
+    private OutlierClassifier outlierClassifier;
+    private Summarizer summarizer;
+
+    public MacrobaseAnalyzer(MacroBaseConf conf) {};
+
+    public AnalysisResult analyze() {
+        for (Datum nextDatum = ingester.getNextDatum(); ; nextDatum = ingester.getNextDatum()) {
```
pbailis: What about using the Java `Iterator<Datum>` interface?

pbailis: Avoids the ugliness of `null` etc.

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
```diff
@@ -0,0 +1,47 @@
...
+            // While there are data points coming.
+            if (nextDatum == null) {
+               break;
+            }
+            // Apply transform(s) to the datum that we got from our engine.
+            Datum transformedDatum = transformation.transform(nextDatum);
+	        // Add transformed datum to the outlier classifier (could also be a
+	        // contextual classifier). This will only update the
+	        // classificationResults once per window, for performance optimization.
+            OutlierClassificationResult classificationResult = outlierClassifier.addDatum(transformedDatum);
```
pbailis: Prefer `classify` to `addDatum`

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
```diff
@@ -0,0 +1,47 @@
...
+            // Summarizer will summarize outliers.
+	        // This will actually perform any computation only if the
+	        // classificationResult has been changed.
+            List<Summary> summaries = summarizer.summarize(classificationResult);
```
pbailis: What about breaking this down into two:

`void observe(Datum d)`

`Summary summarize()`

src/main/java/macrobase/analysis/result/OutlierClassificationResult.java
```diff
@@ -0,0 +1,4 @@
+package macrobase.analysis.result;
+
+public abstract class OutlierClassificationResult {
```
pbailis: Do we think we'll want to overload this a bunch?

Off the top of my head, the basic members we'll almost always want:
 - The datum
 - boolean: isOutlier
 - maybe the score...

pbailis:
Some comments inline...

More generally, can we turn this into a more dataflow-like API?

Right now, control flow is explicit in the while loop. I'm curious if we can replace this via implicit control flow by using iterators: 

```
public abstract class OutlierClassifier {
    public OutlierClassifier(MacroBaseConf conf, Iterator<Datum> input) {}
    public Iterator<OutlierClassificationResult> getResults() {}
}
```

pbailis:
To demonstrate the importance of this iterator model, imagine a wrapper for our existing `FeatureTransformers`:

```
public class BatchFeatureTransformer {
   private List<OutlierClassificationResult> output;
   
    // existing interfaces
    private abstract train(List<Datum>);
    private double score(Datum);    

    public OutlierClassifier(MacroBaseConf conf, Iterator<Datum> input) {
        // drain input to list
        List<Datum> inputList = Lists.newArrayList(input);
        train(inputList);
        for(Datum d : inputList) {
            // score / wrap each datum in a result...
            output.add(score(d));
        }
    }
    public Iterator<OutlierClassificationResult> output() {
        return output.iterator();
    }
}

}

```

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
mamikonyana: Could you elaborate a little more? When would we call each?

src/main/java/macrobase/analysis/result/OutlierClassificationResult.java
mamikonyana: I think we'll probably have just one class implementing this, but I like having an abstract class for each of the components, then we have a clear API that is visible. Also for this stage I didn't seem to have an alternative, so if you feel strongly we can just convert this into a normal class in the future

pbailis:
Another thought: can we just have each of these classes implement appropriate `Iterator<>` interfaces? `FeatureTransformer implements Iterator<Datum>`, `OutlierClassification implements Iterator<OutlierClassificationResult>`?

mamikonyana:
I like the iterator approach too. I'll try to change this to have iterators and be more data flow like

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
pbailis: Sure. The idea is this:

1.) It's expensive to generate summaries. We probably don't want to do so on every tuple. But we do want to incrementally maintain the data structures required.
2.) When we do want to generate the summary, we can do so on demand. So the summarizer is like a sink in the dataflow.

I think we fix this problem by having Summarizer return an `Iterator<Summary>` objects...

pbailis:
@mamikonyana Awesome. If you stub this out, I can take a stab tomorrow. Or you can write it and we can discuss on Monday. What's your preference?

src/main/java/macrobase/analysis/result/OutlierClassificationResult.java
pbailis: I'm in favor of interfaces, but they probably make the most sense if we have a clear set of well-defined functions like `getScore()` `getDatum()` `isOutlier()`, etc. An interface / abstract class without any members or member functions seems slightly gratuitious.

So, what do you think about having an interface `OutlierClassificationResult` that has the above methods?

mamikonyana:
I will probably have something later today, so you can take a look tomorrow. And hopefully others will have some input too. And we can merge this on Monday and start coding on Monday as well?

pbailis:
Sounds good; take a pass and I'll take a look! We should be able to knock out most of this in an afternoon of heavy lifting with support from an IDE.

pbailis:
When you have an update, let's make sure @deepakn94, @viveksjain, and @raininthesun know!

src/main/java/macrobase/analysis/result/OutlierClassificationResult.java
mamikonyana: The reason `OutlierClassificationResult` is empty is that I haven't defined the API yet, I decided to leave that until we finalize the rest of the API. Because based on what we decide for the other things this will change. If we are happy with the rest of API, let's think about how to do this. I agree that this should not be left empty.

I also though of this class/interface as encompassing all the results that classifier would return, so it would probably have methods more like `getOutliers()`, `getInliers()`, or maybe iterators instead of getters.. I'm open to suggestions.

src/main/java/macrobase/analysis/summary/Summary.java
```diff
@@ -0,0 +1,7 @@
+package macrobase.analysis.summary;
+
+/**
+ * Represents a summarization result, currently an itemset with a score and weight.
+ */
+public abstract class Summary {
```
mamikonyana: I'm open to suggestions here too. Summary should have a defined interface, but I'm not sure what exactly is needed from a Summary. `List<Datum> getMembers()`? `getWeight()`?

mamikonyana:
I added bunch of new code. @deepakn94 @pbailis @viveksjain @raininthesun please comment

src/main/java/macrobase/analysis/summary/Summarizer.java
```diff
@@ -0,0 +1,27 @@
...
+import macrobase.analysis.result.AnalysisResult;
+import macrobase.analysis.result.OutlierClassificationResult;
+import macrobase.analysis.summary.itemset.result.ItemsetResult;
+import macrobase.conf.MacroBaseConf;
+import macrobase.datamodel.Datum;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class Summarizer {
```
pbailis: Can this return Iterator<Summary>?

src/main/java/macrobase/analysis/result/OutlierClassificationResult.java
```diff
@@ -0,0 +1,11 @@
+package macrobase.analysis.result;
+
+import macrobase.datamodel.Datum;
+
+import java.util.List;
+
+public abstract class OutlierClassificationResult {
+
+    public abstract List<Datum> getOutliers();
```
pbailis: This seems appropriate for batch but not tuple-at-a-time processing?

Why not just:
 - Datum
 - isOutlier

Where different methods can optionally pass along a "score": a threshold-based method could have the score, a KDE could have a percentile, etc.

src/main/java/macrobase/analysis/result/MacrobaseAnalysisResult.java
```diff
@@ -0,0 +1,26 @@
...
+    private List<Summary> summaries;
+    private OutlierClassificationResult outlierClassificationResult;
+
+    public MacrobaseAnalysisResult(List<Summary> summaries,
+                                   OutlierClassificationResult outlierClassificationResult) {
+        this.summaries = summaries;
+        this.outlierClassificationResult = outlierClassificationResult;
+    };
+
+    public List<Summary> getSummaries() {
```
pbailis: Is there any benefit to just having a single Summary?

pom.xml
```diff
@@ -156,6 +156,11 @@
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
+      <artifactId>commons-collections4</artifactId>
```
pbailis: If we have Guava, do we need commons-collections?

src/main/java/macrobase/analysis/summary/Summary.java
pbailis: Yeah, not sure, also, not sure we need a full List<Summary> below. What if we just base it on `List<ItemsetResult>`?

pbailis:
Lots of small nits.

My main concern: OutlierClassificationResult` should be refactored for tuple-at-a-time processing.

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
```diff
@@ -0,0 +1,55 @@
...
+    // outlierClassifier = constructOutlierClassifier(conf);
+    // Summarizer = new Summarizer(conf, outlierClassifier);
+
+    @Override
+    public MacrobaseAnalysisResult next() {
+
+        int index = 0;
+        for (Datum nextDatum : new IteratorIterable<>(source)) {
+            summarizer.observe(nextDatum);
+            if (index > windowSize) {
```
viveksjain: Is it only possible to summarize over a window? Or can we do incremental summarization that's more in line with the rest of the tuple-at-a-time interfaces?

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
```diff
@@ -0,0 +1,55 @@
...
+    // windowSize = conf.getWindowSize();
+    // outlierClassifier = constructOutlierClassifier(conf);
+    // Summarizer = new Summarizer(conf, outlierClassifier);
+
+    @Override
+    public MacrobaseAnalysisResult next() {
+
+        int index = 0;
+        for (Datum nextDatum : new IteratorIterable<>(source)) {
+            summarizer.observe(nextDatum);
```
viveksjain: It seems to me that it would be cleaner for Summarizer to also be an Iterator. In its constructor, you can pass it the source so that the summarizer can call `source.next` as required. It's unclear what exactly `Summarizer.next` would return (`this`?), but we could just ignore its value here.

pbailis: Yeah, can't `Summarizer` return an `Iterator<Summary>` and then we just return `summarizer.getNext()`?

Adapted from before, something like: `return (new Summarizer(conf, new OutlierClassifier(conf, new BasicDataTransform(conf, new DataIngest(conf).output()).output()).output()).output()).getNext();`

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
```diff
@@ -0,0 +1,19 @@
...
+import macrobase.datamodel.Datum;
+
+public abstract class OutlierClassifier {
+
+    private OutlierClassificationResult lastResult;
+
+    public OutlierClassifier(MacroBaseConf conf) {}
+    public abstract void observe(Datum datumToClassify);
+
+    public abstract OutlierClassificationResult classify();
```
viveksjain: What are the semantics for `classify`? What data is it classifying?

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
pbailis: @viveksjain: If summarization returns an `Iterator<Summary>`, will it solve this problem?

src/main/java/macrobase/transform/FeatureTransform.java
```diff
@@ -0,0 +1,39 @@
...
+        this.ingester = ingester;
+    }
+
+    /**
+      * Applies the transformation this class represents to the datum given and
+      * outputs a new transformed datum.
+      * @param datumToTransform datum that should be transformed
+      * @return new Datum that has transformation applied to it.
+      */
+    public Datum transform(Datum datumToTransform) {
```
viveksjain: Maybe this should be protected.

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
mamikonyana: It is classifying all the data that was observed.

viveksjain:
Looking good so far! I agree with Peter that everything (where possible) should be an iterator.

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
mamikonyana: It could. I don't like that API that much. It's too much nesting.
I'm not sure I am completely sold on the `return (new Summarizer(conf, new OutlierClassifier(conf, new BasicDataTransform(conf, new DataIngest(conf).output()).output()).output()).output()).getNext();` line.
I can't imagine where in the process would the windowing happen?
Also a method returning an Iterator seems ugly to me, but if that's a standard java thing, I'd be ok with that.

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
pbailis: I think that if OutlierClassificationResult has both `Datum getDatum()` and `boolean isOutlier()` functions, this will be clearer.

src/main/java/macrobase/transform/FeatureTransform.java
pbailis: What's the rationale? 

Say I want to link to MacroBase and implement my own `FeatureTransform`. Do I need to place my `FeatureTransform` in the `macrobase.transform` package?

(It's possible I am misunderstanding; it's late :v:.)

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
mamikonyana: What do you mean by incremental summarization?
In this design Summary is just one summary, and Summarizer returns a list of Summaries (itemsets).
We could change such that Summary is an ensemble of itemsets.. I don't really see the point of doing that though

src/main/java/macrobase/analysis/summary/Summary.java
mamikonyana: We could, I just wanted new abstract classes, not to carry anything old with it. We can rename things once the API is finalized.

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
viveksjain: Oh I see, so the windowing would just be handled inside `Summarizer.next` rather than here? That makes sense. Although perhaps a little inconsistent since all the other iterators return/operate over a single datum.

mamikonyana:
I think I'm a little unclear on something. How does a streaming OutlierDetector work? It classifies a given datum as outlier/inlier? or based on previous results it outputs set of outliers/inliers on demand (like the summarizer)?

pbailis:
@mamikonyana Each putlier detection module has two jobs: 
 1. keep the model up to date
 2. score points

In tuple-at-a-time data processing, you need to do both at once. So, when we see a new data point, we need to:
 1. update the model based on the new point
 2. score the new point

Both can be accomplished via a single iteration call...

src/main/java/macrobase/transform/FeatureTransform.java
viveksjain: If you subclass `FeatureTransform` you can override a protected method. The main reason is to make sure clients don't call transform manually and instead use the public facing iterator APIs. (It's partly because I was wondering why transform existed if `FeatureTransform implements Iterator` until I saw the implementation of next. Although that could also be somewhat cleared up by just moving next to be above transform :smile:).

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
pbailis: Yes, that's the idea. I agree this is inconsistent compared to the other operators, but, insofar as a `Summarizer` returns a `Summary`, it seems a somewhat natural departure.

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
viveksjain: So what is the difference between `classify` and `getResult`? And if it classifies all of the data does that mean it has to keep the whole input stream around? That seems like a bad idea for high volume streams.

src/main/java/macrobase/transform/FeatureTransform.java
pbailis: I'm with you until "although that could also be somewhat cleared up by just moving next to be above transform".

`protected` sounds good to me. @mamikonyana?

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
pbailis: @mamikonyana: windowing would happen internally to a `FeatureTransform`...

It's pretty clean if you break it up across lines:

```
DataIngest ingest = new DataIngest(conf);
DataTransform transform = new DataTransform(conf, ingest);
OutlierClassifier c = new OutlierClassifier(conf, transform);
Summarizer s = new Summarizer(conf, c);

return s.getNext();
```

Otherwise, why have the explicit `for` loop?

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
pbailis: The proposal:
  * `classify()` returns an `OutlierClassificationResult`
  * `OutlierClassificationResult` implements a method `boolean isOutlier()` and a method `Datum getDatum()`

You don't need to keep the entire stream around to implement this.

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
viveksjain: Actually, ignore the incremental summarization. To be clearer, for this and my comment on the above line, I was thinking of making `Summarizer` an iterator. So like most of the other classes, you could then call next on it and it would automatically fetch the next input from the classifier so that this explicit call to `observe` isn't required.

So the new pipeline is Ingest -> Feature extraction -> Context selection -> Density estimation -> Summarization. Once you've constructed the pipeline, the idea is just to call `next` on the last step of the pipeline, and each step calls `next` on the previous step as required. I think that makes for a clean and modular architecture, both conceptually and code-wise.

pbailis: +1! I think Summarizer should be an iterator as well.

"Iterators all the way down" is a classic design pattern in database engines dating back to the early days...

pom.xml
mamikonyana: I have no clue. I know that [IteratorIterable](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/iterators/IteratorIterable.html) is in colletions4, and it didn't seem to compile without this.

mamikonyana:
questions: 
- what's the purpose of `MacrobaseAnalyser`? Should that be an Iterator too?
Currently it is, but what do we want it to output? As of the current design we have it output best summaries together with performance metrics.

- Summerizer should not stick to the tuple-at-a-time strategy, correct? It should produce summaries when it has a new summary? The way I imagine summary is that it's a new insight that could be potentially shown to the user. With current streaming approach this is batch based and returns a set of summaries every window size, right? (when we merge FP Trees) this was the reason I had `public List<Summary> getSummaries()`. So Summary should contain a list of itemsets?

I've added comments to some of the classes, let me know if any of them does not sound correct.

viveksjain:
From my understanding (may be incorrect):
 - `MacrobaseAnalyser` is basically the glue code that sets everything up and gets it running. Outputting summaries and performance metrics sounds good to me.
 - Yes, when  you call `Summarizer.next` it will actually consume a window size number of tuples. So in that case I don't think we need a separate `Summarizer.summarize` function, and instead we just call `Summarizer.next`. I'm not sure about `Summary` vs `ItemSet` - I defer to @pbailis on that.

pbailis:
> - what's the purpose of `MacrobaseAnalyser`? Should that be an Iterator too?
 > Currently it is, but what do we want it to output? As of the current design we have it output best summaries together with performance metrics.

The (current) purpose is that it implements a generic pipeline of operators. I imagine that different users could implement different pipelines if they wanted...

I think it's fine to make it an iterator.

 > - Summerizer should not stick to the tuple-at-a-time strategy, correct? It should produce summaries when it has a new summary? The way I imagine summary is that it's a new insight that could be potentially shown to the user.

The way I think of the summarizer:

The summarizer takes in tuples by calling `next()` on the input (classified points).
When someone calls `next()` on the summarizer, the summarizer returns the "latest" summary.

So, we're implementing is a pull-based iterator model: each module produces more output when you call `next()`. It's possible the return value has been pre-computed, but you have to ask for the `next` item in order to get it.

You can turn a "pull" into a "push" by having a separate operator that basically polls on the output (a common term for this is called an "exchange" operator, which a system called Volcano pioneered in the 1990s).

> With current streaming approach this is batch based and returns a set of summaries every window size, right? 

Not exactly. At window boundaries, we maintain the FP tree. But we only generate the summaries when we call `getItemsets()`. My proposal is that the summarizer implement `Iterator<Summary>` to replace this behavior.

 > this was the reason I had `public List<Summary> getSummaries()`. So Summary should contain a list of itemsets?

Yeah, let's have each `Summary` contain a `List<Itemset>`.

> I've added comments to some of the classes, let me know if any of them does not sound correct.

Will look.

pom.xml
pbailis: Got it. No problem. Apparently the Guava folks don't like this, but I do: https://github.com/google/guava/issues/796 :)

src/main/java/macrobase/analysis/MacrobaseAnalyzer.java
```diff
@@ -0,0 +1,42 @@
...
+import macrobase.ingest.DataIngester;
+import macrobase.transform.FeatureTransform;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
+
+import java.util.Iterator;
+
+public class MacrobaseAnalyzer implements Iterator<MacrobaseAnalysisResult> {
+
+    private DataIngester ingester;
+    private FeatureTransform source;
```
pbailis: slight preference: source -> transform

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Consumes a tuple at-a-time from a FeatureTransform and for each tuple outputs
+ * a OutlierClassificationResult that should indicate if the tuple observed was an
+ * outlier or not.
+ */
+public abstract class OutlierClassifier implements Iterator<OutlierClassificationResult> {
+
+    private FeatureTransform featureTransform;
+
+    public OutlierClassifier(MacroBaseConf conf, FeatureTransform transform) {
```
pbailis: Why does this take a FeatureTransform and not an Iterable<Datum>?

src/main/java/macrobase/transform/FeatureTransform.java
```diff
@@ -0,0 +1,39 @@
+package macrobase.transform;
+
+import macrobase.conf.MacroBaseConf;
+import macrobase.datamodel.Datum;
+import macrobase.ingest.DataIngester;
+
+import javax.xml.crypto.Data;
```
pbailis: Import probably not what you wanted :)

src/main/java/macrobase/transform/FeatureTransform.java
```diff
@@ -0,0 +1,39 @@
...
+        this.ingester = ingester;
+    }
+
+    /**
+      * Applies the transformation this class represents to the datum given and
+      * outputs a new transformed datum.
+      * @param datumToTransform datum that should be transformed
+      * @return new Datum that has transformation applied to it.
+      */
+    protected Datum transform(Datum datumToTransform) {
```
pbailis: Should this be abstract?

pbailis:
Looking very reasonable...

pbailis:
@viveksjain 

 > MacrobaseAnalyser is basically the glue code that sets everything up and gets it running. Outputting summaries and performance metrics sounds good to me.

Agreed. The classes that extend MacroBaseAnalyzer contain some canned analyses. Eventually users should be able to write their own pipelines.

 > Yes, when you call Summarizer.next it will actually consume a window size number of tuples. So in that case I don't think we need a separate Summarizer.summarize function, and instead we just call Summarizer.next.

+1!

> I'm not sure about Summary vs ItemSet - I defer to @pbailis on that.

I think a Summary should contain potentially many itemsets.

src/main/java/macrobase/analysis/outlier/OutlierClassifier.java
mamikonyana: `Iterable` or `Iterator`?

src/main/java/macrobase/transform/FeatureTransform.java
mamikonyana: no, not exactly :)

mamikonyana:
I made the suggested changes
- `source` -> `dataTransform`
- Change `OutlierClassifier` to accept `Iterator<Datum>` instead of `DataTransform`

I already have `Summary` containing multiple itemsets.

So is this it?

src/main/java/macrobase/transform/FeatureTransform.java
mamikonyana: It seems like you can't have abstract private/protected methods, I added so that it throws `throw new NotImplementedException()`

viveksjain: According to http://stackoverflow.com/questions/3851747/java-how-to-implement-private-abstract-methods you can have `protected abstract` but not `private abstract`.

mamikonyana: fixed.

pbailis:
@viveksjain @deepakn94 @raininthesun: this has a ways to go in terms of test coverage and supporting some of Xu's features, but it runs...

Test/bug fixes go in tomorrow. Curious what you think:
https://github.com/stanford-futuredata/macrobase/blob/newDesign/src/main/java/macrobase/analysis/OnePassStreamingAnalyzer.java#L34
https://github.com/stanford-futuredata/macrobase/blob/newDesign/src/main/java/macrobase/analysis/BatchAnalyzer.java#L60

pbailis:
Can/should we have a ContextualDatum that subclasses Datum, like TimeDatum does?

pbailis:
We can have multiple inheritance of interfaces. So, what if we make contextual- and time-based interfaces? @mamikonyana @viveksjain

raininthesun:
> Can/should we have a ContextualDatum that subclasses Datum, like TimeDatum does?

From my understanding, inheritance is mainly used for achieve polymorphism, such as sub-class should override the super-class method. For example, different Shape subclasses should override the area() method of a superclass Shape to calculate the area of different shapes. 

In our case, TimeDatum or ContextualDatum doesn't really need to override any method of Datum, but rather only add additional information. Does this mean that we should simply have one Datum class, and provides multiple constructor classes.

not entirely sure which one is better.

src/main/java/macrobase/analysis/BatchAnalyzer.java
```diff
@@ -63,87 +60,19 @@ public AnalysisResult analyze() throws SQLException, IOException, ConfigurationE
...
-        log.debug("Tuples / second = {} tuples / second", tuplesPerSecond);
+        Summarizer summarizer = new BatchSummarizer(conf, classifier);
 
-        tuplesPerSecond = ((double) data.size()) / ((double) tsw2.elapsed(TimeUnit.MICROSECONDS));
-        tuplesPerSecond *= 1000000;
-        log.debug("Tuples / second w/o itemset mining = {} tuples / second", tuplesPerSecond);
+        Summary result = summarizer.next();
 
-        return new AnalysisResult(outlierSize, inlierSize, loadTime, classifyTime, summarizeTime, isr);
+        return new AnalysisResult(0, 0, 0, 0, 0, result.getItemsets());
```
deepakn94: Would be nice to have a TODO here to make sure we eventually don't just return 0s. (Better yet, some code to actually return the right values)

src/main/java/macrobase/analysis/BatchAnalyzer.java
```diff
@@ -1,32 +1,31 @@
...
 import org.slf4j.LoggerFactory;
 
 import java.io.*;
 import java.sql.SQLException;
-import java.util.*;
-import java.util.concurrent.TimeUnit;
+import java.util.List;
+import java.util.Map;
 
 public class BatchAnalyzer extends BaseAnalyzer {
```
deepakn94: What's the difference between `BaseAnalyzer` and `MacrobaseAnalyzer`?

deepakn94:
I'll take a deeper look at this tomorrow, but looks good so far!

viveksjain:
> In our case, TimeDatum or ContextualDatum doesn't really need to override any method of Datum, but rather only add additional information. Does this mean that we should simply have one Datum class, and provides multiple constructor methods.

Yeah, I think we had that initially but then I thought it would be cleaner (and require one fewer pointer per Datum) if there was a separate Datum and TimeDatum. But if we want to mix multiple optional items I would favor composition over inheritance, so store instance variables containing time and whatever else we need. I'm not even sure how storing it in interfaces would work (and I suspect it would end up requiring lots of casts).

src/main/java/macrobase/analysis/BatchAnalyzer.java
```diff
@@ -63,87 +60,19 @@ public AnalysisResult analyze() throws SQLException, IOException, ConfigurationE
     	if(contextualEnabled){
     		contextualAnalyze();
```
viveksjain: Is this only to print, or is the output of this used further down the pipeline (I don't know how context works)?

pbailis: > (I don't know how context works)

Neither do we, really, so we're playing it safe for now... :)

pbailis:
The general consensus seems to be that we should allow Datum to contain a bunch of variables.

This is a-okay by me; you've convinced me. The main remaining concern I have is in extensibility to other ingest sources. However, an intrepid soul can probably come up with a new base class for doing so....

pbailis:
@viveksjain @raininthesun @deepakn94 
This is ready for merge. If one of you wants to take a look, please go ahead! Alternatively, Arsen and I wrote almost all of this together, so we could also look it over another time and then merge ourselves.

deepakn94:
`Files changed: 124`
This is going to go well.

deepakn94:
I like how everything was squashed into one commit. :P Don't want to contaminate the final `git log` with all those nasty commits.

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public BatchingPercentileClassifier(MacroBaseConf conf, Iterator<Datum> input, double percentile) {
+        super(conf, input);
+        this.targetPercentile = percentile;
+    }
+
+    @Override
+    public OutlierClassificationResult next() {
+        if(resultIterator == null) {
```
deepakn94: How do you envision using this `BatchingPercentileClassifier` in the pipeline? Would you create a new instance of this class everytime you want a new set of results?

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private final Periodic reservoirDecayer;
+    private final Periodic percentileUpdater;
+
+    List<OutlierClassificationResult> warmupOutput = new ArrayList<>();
+
+    public EWAppxPercentileOutlierClassifier(MacroBaseConf conf, Iterator<Datum> input) {
+        this(conf,
+             input,
+             conf.getInt(MacroBaseConf.SCORE_RESERVOIR_SIZE, MacroBaseDefaults.SCORE_RESERVOIR_SIZE),
```
deepakn94: Ugh, this is slightly ugly.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
```diff
@@ -0,0 +1,104 @@
...
+        return input.hasNext();
+    }
+
+    @Override
+    public OutlierClassificationResult next() {
+        if(!warmupOutput.isEmpty()) {
+            return warmupOutput.remove(0);
+        }
+
+        reservoirDecayer.update();
```
deepakn94: I presume these aren't actually decaying / updating percentiles every tuple. If this is indeed the case, consider renaming to `updateIfNecessary()`.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
```diff
@@ -0,0 +1,104 @@
...
+        this.percentileUpdater = new Periodic(updatePeriodType,
+                                              updatePeriod,
+                                              () -> updateThreshold(percentile));
+
+        this.reservoirDecayer = new Periodic(decayPeriodType,
+                                             decayPeriod,
+                                             reservoir::advancePeriod);
+
+        List<DatumWithNorm> warmupInput = new ArrayList<>(warmupCount);
+        for(int i = 0; i < warmupCount; ++i) {
```
deepakn94: Not a huge fan of handling warmup tuples in the constructor.

src/main/java/macrobase/analysis/classify/StaticThresholdClassifier.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public boolean hasNext() {
+        return input.hasNext();
+    }
+
+    @Override
+    public OutlierClassificationResult next() {
+        Datum nextInput = input.next();
+        return new OutlierClassificationResult(nextInput, thresholdSquared > nextInput.getMetrics().getNorm());
```
deepakn94: `Datum`s can have scores too? Is there a distinction between the raw Datum that we get from the Postgres instance, and these Datums? If so, consider re-naming to make that distinction more clear.

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
```diff
@@ -0,0 +1,89 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class BasicBatchedPipeline extends OneShotPipeline {
```
deepakn94: `OneShotPipeline`?!

src/main/java/macrobase/analysis/pipeline/OneShotPipeline.java
```diff
@@ -0,0 +1,36 @@
...
+
+import macrobase.analysis.result.AnalysisResult;
+import macrobase.conf.ConfigurationException;
+import macrobase.conf.MacroBaseConf;
+
+import java.io.IOException;
+import java.sql.SQLException;
+
+
+abstract public class OneShotPipeline extends AbstractPipeline {
```
deepakn94: Ugh, this is ugly. Not a huge fan of returning an Iterator if we only get results only once.

src/main/java/macrobase/analysis/pipeline/BasicOneShotEWStreamingPipeline.java
```diff
@@ -0,0 +1,56 @@
...
+        FeatureTransform featureTransform = new EWFeatureTransform(conf, batchIngest);
+        OutlierClassifier outlierClassifier = new EWAppxPercentileOutlierClassifier(conf, featureTransform);
+        Summarizer summarizer = new EWStreamingSummarizer(conf, outlierClassifier);
+        Summary result = summarizer.next();
+
+        final long endMs = System.currentTimeMillis();
+        final long loadMs = batchIngest.getFinishTimeMs() - startMs;
+        final long totalMs = endMs - batchIngest.getFinishTimeMs();
+        final long summarizeMs = result.getCreationTimeMs();
+        final long executeMs = totalMs - summarizeMs;
```
deepakn94: Huge fan of this easy timing :100:

src/main/java/macrobase/analysis/result/DatumWithNorm.java
```diff
@@ -0,0 +1,26 @@
...
+        return norm;
+    }
+
+    public Datum getDatum() {
+        return datum;
+    }
+
+    public DatumWithNorm(Datum d) {
+        this.datum = d;
+        this.norm = d.getMetrics().getNorm();
```
deepakn94: What's the point of DatumWithNorm if the norm is already available in a standard Datum?

src/main/java/macrobase/analysis/summary/Summarizer.java
```diff
@@ -0,0 +1,29 @@
...
+ # when there has been an update.
+ */
+public abstract class Summarizer implements Iterator<Summary> {
+
+    protected final Iterator<OutlierClassificationResult> input;
+    protected final DatumEncoder encoder;
+
+    public
+    Summarizer(MacroBaseConf conf, Iterator<OutlierClassificationResult> classificationResults) {
+        input = classificationResults;
```
deepakn94: Nit: Let's keep the naming uniform here: the parameter name can just be input.

src/main/java/macrobase/analysis/summary/BatchSummarizer.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class BatchSummarizer extends Summarizer {
+    private FPGrowthEmerging fpg;
+    protected final Double minSupport;
+    protected final Double minOIRatio;
+
+    public BatchSummarizer(MacroBaseConf conf, Iterator<OutlierClassificationResult> results) {
```
deepakn94: Let's rename `results` to `input` here.
The reason I'm being pedantic about this is that we have an `input` in the `next()` function and it's not immediately obvious where that comes from.

src/main/java/macrobase/analysis/summary/result/DatumWithScore.java
```diff
@@ -2,6 +2,7 @@
 
 import macrobase.datamodel.Datum;
 
+// TODO: REMOVE
```
deepakn94: Why can't we remove this now?

src/main/java/macrobase/analysis/transform/EWFeatureTransform.java
```diff
@@ -0,0 +1,97 @@
...
+        decayer = new Periodic(decayPeriodType,
+                               decayPeriod,
+                               reservoir::advancePeriod);
+
+        retrainer = new Periodic(trainingPeriodType,
+                                 trainingPeriod,
+                                 () -> scorer.train(reservoir.getReservoir()));
+
+        List<Datum> warmupInput = new ArrayList<>(warmupCount);
+        for(int i = 0; i < warmupCount; ++i) {
```
deepakn94: Same comments as before with regards to handling warmup tuples in the constructor.

deepakn94: What would we do in an actual streaming setting where all tuples aren't available a-priori.

src/main/java/macrobase/conf/MacroBaseConf.java
```diff
@@ -92,18 +118,63 @@
...
+                if (metricsDimensions == 1) {
+                    log.info("By default: using MAD detector for dimension 1 metric.");
+                    return new MAD(this);
+                } else {
+                    log.info("By default: using MCD detector for dimension {} metrics.", metricsDimensions);
+                    MinCovDet ret = new MinCovDet(this);
+                    return ret;
+                }
+            case MAD:
+                log.info("Using MAD detector.");
```
deepakn94: If we're calling them transformers, detector->transformer.

src/test/java/macrobase/pipeline/BasicBatchedPipelineTest.java
```diff
@@ -18,8 +16,8 @@
 
 import static org.junit.Assert.*;
 
-public class BatchAnalyzerTest {
-    private static final Logger log = LoggerFactory.getLogger(BatchAnalyzerTest.class);
+public class BasicBatchedPipelineTest {
```
deepakn94: Could be useful to have a short comment describing what this test is trying to test. (Relevant for most testing files, but especially the end-to-end ones)

deepakn94:
Wow, that was a lot of code. I had mostly minor comments, some higher-level design questions.
Overall, looks good. Ping me when you've gone through the comments and I can take another look.

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
mamikonyana: You mean if we wanted to run two experiments with the same outlier detection algorithms, but different percentile scores?
Or you just want to change the threshold from pipeline to pipeline?

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
mamikonyana: suggestion to improve?

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
mamikonyana: I'm not sure. @pbailis knows this part

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
mamikonyana: this is general strategy we used. I think we also drain up iterators in constructors in general too. You have a better design suggestion where to handle this? We just end up with cleaner code by doing this rather than checking if it needs to be done from inside `.next()`

src/main/java/macrobase/analysis/classify/StaticThresholdClassifier.java
mamikonyana: no real distinction, the only distinction is that this datum is a Datum that had gone through the feature transformation, so the values in the metrics are not the values that we got out of postgres, rather the values that we gave them based on transformations applied, e.g. windowed average, Zscore, inverse of the value, etc. Whatever we have defined in the pipeline.

src/main/java/macrobase/analysis/pipeline/OneShotPipeline.java
mamikonyana: Agreed. But, we had a dilemma how to handle returning of results during the stream processing. Ideally we would want to return results multiple times, but we don't have user interface to support that.. also we are not sure if we should return updated results or the results about new data points.. So this seemed like a solution that works for now.

src/main/java/macrobase/analysis/result/DatumWithNorm.java
mamikonyana: not sure, @pbailis ?

src/main/java/macrobase/analysis/summary/Summarizer.java
mamikonyana: cool.

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
deepakn94: Sure, what happens with those two use cases?
What happens if I call next() to get results once, then I want to wait for more input tuples to filter into the system, and then want results again.

src/main/java/macrobase/analysis/summary/BatchSummarizer.java
mamikonyana: done.

src/main/java/macrobase/analysis/summary/result/DatumWithScore.java
mamikonyana: nice catch.

mamikonyana: we've got a problem that contextual Outlier Detection is still using this.. So we'll have to keep it as a TODO for now.

src/main/java/macrobase/conf/MacroBaseConf.java
mamikonyana: done.

src/test/java/macrobase/pipeline/BasicBatchedPipelineTest.java
mamikonyana: @pbailis ?

mamikonyana:
Replied to most of the comments, and made some of the changes.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
deepakn94: Can't think of anything off the top of my head. Hard to have a clean solutions since we can have arbitrary sequences of operators now.

src/main/java/macrobase/analysis/pipeline/OneShotPipeline.java
deepakn94: I wonder if we could have the final stage of the pipeline be something that doesn't sub-class Iterator, but accumulates results instead into some internal state. Then we could support multiple functions on it such as `getNewResults`, `getAllResults`, etc. Seems more malleable then the iterator's `next` approach.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
deepakn94: I mention this in a later comment, but here's why I take issue with handling this in the constructor: What would we do in an actual streaming setting where all tuples aren't available a-priori? Would we have to wait till `warmupTuples` pass through the pipeline before being able to instantiate this class? That seems non-ideal.

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
pbailis: @deepakn94 This is a convenience wrapper for the existing batch functionality. Is your proposal that we should check the input for more tuples to drain?

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
pbailis: Fine for now.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
pbailis: Well, they're doing arbitrary computation. Maybe just `runIfNecessary`.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
pbailis: > Would we have to wait till warmupTuples pass through the pipeline before being able to instantiate this class?

Yes. This is the idea of a warmup period. There's no output until `warmupTuples` have passed.

src/main/java/macrobase/analysis/classify/StaticThresholdClassifier.java
pbailis: > If so, consider re-naming to make that distinction more clear.

A core goal of this refactor is to get rid of the idea of a "score" at all. Metrics are all that classifiers operate over now.

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
pbailis: Meaning: runs once, compared to a truly streaming pipeline.

src/main/java/macrobase/analysis/pipeline/OneShotPipeline.java
pbailis: > Seems more malleable then the iterator's next approach.

I agree it'd be convenient, but the goal is to make this iterators all the way down, independent of internal batching, etc.

src/main/java/macrobase/analysis/pipeline/BasicOneShotEWStreamingPipeline.java
pbailis: An alternative is to have `TimedBatchIngest` return a started timer. Seemed less clean.

src/main/java/macrobase/analysis/result/DatumWithNorm.java
pbailis: This is a utility class; `norm` is cached.

src/main/java/macrobase/analysis/summary/result/DatumWithScore.java
pbailis: Let's still update the TODO.

src/main/java/macrobase/analysis/transform/EWFeatureTransform.java
pbailis: See response above.

src/test/java/macrobase/pipeline/BasicBatchedPipelineTest.java
pbailis: There's some commenting already, but we could do better: https://github.com/stanford-futuredata/macrobase/blob/newDesign/src/test/java/macrobase/pipeline/BasicBatchedPipelineTest.java#L23

pbailis:
@mamikonyana Looks like after we change `update()` to `runIfNecessary()`, this is good to merge.

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
deepakn94: I'm not suggesting anything here. I'm asking how the existing code would be used in those 3 use cases.

src/main/java/macrobase/analysis/classify/EWAppxPercentileOutlierClassifier.java
deepakn94: Well, it just seems weird to have arbitrary stalling within a constructor.

src/main/java/macrobase/analysis/classify/StaticThresholdClassifier.java
deepakn94: I see.

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
deepakn94: Yeah, I got that.

src/main/java/macrobase/analysis/pipeline/OneShotPipeline.java
deepakn94: I'm suggesting having a wrapper around the entire pipeline do this for us. The main difference from the current approach is that the wrapper would keep track of the resultset last produced so that it could compute a diff, return the entire new result set, etc.

mamikonyana:
done!

PR #99, opened 2016-03-22
Add Bayesian Normal Density estimator

mamikonyana:


mamikonyana:
If you want theoretical background, this is a good overview of what's happening in 1 dimensional case (section 3) - http://www.cs.berkeley.edu/~jordan/courses/260-spring10/lectures/lecture5.pdf

This are just results for multivariate case - http://www.fil.ion.ucl.ac.uk/~wpenny/publications/bmn.pdf

pbailis:
So, just to check my understanding: this fits a single Normal distribution using Bayesian inference?

mamikonyana:
Yes that's correct

src/main/java/macrobase/analysis/stats/BayesianNormalDensity.java
```diff
@@ -0,0 +1,85 @@
...
+import macrobase.conf.MacroBaseConf;
+import macrobase.datamodel.Datum;
+import org.apache.commons.math3.linear.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+
+public class BayesianNormalDensity extends BatchTrainScore {
```
pbailis: Comment explaining class would be helpful.

src/main/java/macrobase/analysis/stats/distribution/MultivariateTDistribution.java
```diff
@@ -0,0 +1,60 @@
...
+
+        double determinant = new LUDecomposition(covarianceMatrix).getDeterminant();
+
+        this.multiplier = halfGammaRatio(dimensions + degreesOfFreedom, degreesOfFreedom)/
+                Math.pow(Math.PI * degreesOfFreedom, 0.5 * dimensions) /
+                Math.pow(determinant, 0.5);
+    }
+
+
+    public double density(RealVector vector) {
```
pbailis: Do we want to have a more generic distribution class?

pbailis:
Minor comments; otherwise, LGTM to merge.

src/main/java/macrobase/analysis/stats/distribution/MultivariateTDistribution.java
mamikonyana: They have too many methods in the abstract class from apache, so let's keep it simple for now.

src/main/java/macrobase/analysis/stats/BayesianNormalDensity.java
mamikonyana: adding

PR #107, opened 2016-04-18
Adding local csv energy test data

edgan8:
Will be useful to have a csv file with real data for local testing
w/o the CMT data or a populated postgres db. Also created a conf file
to try using that data.

For that purpose I took the us_tmy3_energy data, performed a 0.1% 
bernoulli tablesample, and joined with building_name since building_id
is the encoded crossproduct of both area and building name.

Since the real data has missing column values, I needed to change
the csv ingester to filter out bad rows by caching all the successfully
parsed rows in memory. This will hopefully not be an issue for any of
the small tests we are using csvparser for, and should be faster
to iterate through.

To run the new energy data:

./bin/batch.sh conf/local_batch.yaml

RefBldgLargeOfficeNew2004 should show up as an outlier attribute


src/test/java/macrobase/ingest/CSVIngesterTest.java
```diff
@@ -1,26 +1,63 @@
...
+import macrobase.conf.MacroBaseConf;
+import macrobase.datamodel.Datum;
 
-import java.io.IOException;
-import java.util.ArrayList;
 
 public class CSVIngesterTest {
     private static final Logger log = LoggerFactory.getLogger(CSVIngesterTest.class);
 
-    public void simpleTest() throws IOException, ConfigurationException {
```
edgan8: @mamikonyana how did you run this test previously? I made it into a JUnit test to run it.

edgan8:
Not sure how to mark people as reviewers on github. @pbailis @viveksjain

src/test/java/macrobase/ingest/CSVIngesterTest.java
pbailis: May not have ran :)

src/main/java/macrobase/ingest/CSVIngester.java
```diff
@@ -50,29 +56,37 @@ public CSVIngester(MacroBaseConf conf) throws ConfigurationException, IOExceptio
...
+        // Load all records into memory to filter out rows with missing data
+        Iterator<CSVRecord> rawIterator = csvParser.iterator();
+        cachedData = new ArrayList<>(1000);
+
+        while (rawIterator.hasNext()) {
+            try {
+                CSVRecord record = rawIterator.next();
+                Datum curRow = parseRecord(record);
+                cachedData.add(curRow);
+            } catch (NumberFormatException e) {
```
pbailis: Do we want to log any errors?

pbailis:
LGTM, though it'd be good to log some error if we do encounter bad rows.
Also, can you compress the file as a `csv.gz`?

edgan8:
Thanks, logged how many bad rows we encounter and switched to gzipped csv.  On real data it's something some fraction of missing data will be expected I think so it's not an outright error, but worth knowing what the fraction is.

(Commit was amended and force pushed so there is still only one commit on this PR, b/c trying to figure out how to have multiple dependent PR)

pbailis:
:100: --

 > (Commit was amended and force pushed so there is still only one commit on this PR, b/c trying to figure out how to have multiple dependent PR)

This is how we do it... :)

PR #108, opened 2016-04-19
Variational Inference for Dirichlet Process Gaussian Mixture

mamikonyana:
This also has some helper classes for diagnostics, I could separate into a differnet pull requets if that will make it easier to go over the diff.

The main thing that this diff accomplishes is that now we can infer an infinite mixture of Gaussians when doing density estimation. It's slow, but it's a good start.

### DIFF
--- a/src/main/java/macrobase/analysis/stats/mixture/VariationalDPMG.java
+++ b/src/main/java/macrobase/analysis/stats/mixture/VariationalDPMG.java
@@ -1,0 +1,288 @@
+package macrobase.analysis.stats.mixture;
+
+import macrobase.analysis.stats.BatchMixtureModel;
+import macrobase.analysis.stats.distribution.MultivariateTDistribution;
+import macrobase.conf.MacroBaseConf;
+import macrobase.conf.MacroBaseDefaults;
+import macrobase.datamodel.Datum;
+import macrobase.util.AlgebraUtils;
+import org.apache.commons.math3.linear.*;
+import org.apache.commons.math3.special.Gamma;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Variational Dirichlet Process Mixture of Gaussians.
+ */
+public class VariationalDPMG extends BatchMixtureModel {
+    private static final Logger log = LoggerFactory.getLogger(VariationalDPMG.class);
+    private final int maxIterationsToConverge;
+
+    // Number of truncated clusters.
+    private int T;
+    // Concentration parameter for the Dirichlet distribution.
+    private double concentrationParameter;
+    private double progressCutoff;
+
+    // Variables governing atoms.
+    // Omega and atomNu for Wishart distribution for the precision matrix of the clusters.
+    private double atomNu[];
+    private List<RealMatrix> atomOmega;
+    // Parameters for Normal distribution for atom position, N(atomLocation, (atomBeta * Lambda))
+    // where Lambda is Wishart distributed given parameters above.
+    private double atomBeta[];
+    private List<RealVector> atomLocation;
+
+    // Parameters describing stick lengths, i.e. shape parameters of Beta distributions.
+    private double shapeParams[][];
+
+    // Parameters for multinomial distributions for data points. i.e. density corresponding to each cluster (atom).
+    private double[][] r;
+
+    // Parameters for Base Distribution of the DP process, which is Wishart-Gaussian
+    private double baseNu;
+    private RealMatrix inverseBaseOmega;  // Use inverse of baseOmega, since it is used in the update equations.
+    private double baseBeta;
+    private RealVector baseLocation;
+
+    private List<MultivariateTDistribution> predictiveDistributions;
+
+    private void updatePredictiveDistributions() {
+        int dimension = atomLocation.get(0).getDimension();
+
+        predictiveDistributions = new ArrayList<>(T);
+        for (int t = 0; t < T; t++) {
+            double scale = (atomNu[t] + 1 - dimension) * atomBeta[t] / (1 + atomBeta[t]);
+            RealMatrix ll = AlgebraUtils.invertMatrix(atomOmega.get(t).scalarMultiply(scale));
+            // TODO: MultivariateTDistribution should support real values for 3rd parameters
+            predictiveDistributions.add(new MultivariateTDistribution(atomLocation.get(t), ll, (int) (atomNu[t] + 1 - dimension)));
+        }
+    }
+
+    public VariationalDPMG(MacroBaseConf conf) {
+        super(conf);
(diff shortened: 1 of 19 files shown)


mamikonyana:
@pbailis can we change the coverage check to be slightly more tolerable?

src/main/java/macrobase/diagnostics/ScoreDumper.java
```diff
@@ -0,0 +1,41 @@
+package macrobase.diagnostics;
```
viveksjain: This file caught my eye - is there a particular reason this couldn't be integrated with src/test/java/macrobase/diagnostic/tasks/ScoreDumpDiagnostic.java?

mamikonyana: yes actually. I had some code that would run diagnostics not from tests. I still want to, because sometimes all i need is run macrobase on a given dataset and dump the results in a file to be able to plot. Currently my solution is to write a test for that case, which does not give me freedom to automatically iterate over configuration parameters or datasets. So I think there should be a diagnostics pipeline or switch in the main package, I can't decide which one would be nicer and more maintainable.

mamikonyana: Also thanks for looking through the code :)

pbailis:
> can we change the coverage check to be slightly more tolerable?

What do you mean?

src/main/java/macrobase/analysis/stats/mixture/VariationalDPMG.java
```diff
@@ -0,0 +1,288 @@
...
+            double oldLogLikelihood = logLikelihood;
+            logLikelihood = 0;
+            for (int n = 0; n < N; n++) {
+                logLikelihood += Math.log(score(data.get(n)));
+            }
+
+            log.info("log likelihood after iteration {} is {}", iteration, logLikelihood);
+
+            if (iteration >= maxIterationsToConverge) {
+                log.info("Breaking because have already run {} iterations", iteration);
```
pbailis: Shouldn't these be log.debug or log.trace?

src/main/java/macrobase/analysis/stats/mixture/VariationalDPMG.java
```diff
@@ -0,0 +1,288 @@
...
+
+            updatePredictiveDistributions();
+
+            double oldLogLikelihood = logLikelihood;
+            logLikelihood = 0;
+            for (int n = 0; n < N; n++) {
+                logLikelihood += Math.log(score(data.get(n)));
+            }
+
+            log.info("log likelihood after iteration {} is {}", iteration, logLikelihood);
```
pbailis: Here too.

src/main/java/macrobase/analysis/stats/mixture/VariationalDPMG.java
```diff
@@ -0,0 +1,288 @@
...
+
+            if (iteration >= maxIterationsToConverge) {
+                log.info("Breaking because have already run {} iterations", iteration);
+                break;
+            }
+
+            double improvement = (logLikelihood - oldLogLikelihood) / ( -logLikelihood);
+            if (improvement >= 0 && improvement < this.progressCutoff) {
+                log.debug("Breaking because improvement was {} percent", improvement * 100);
```
pbailis: Here too.

src/main/java/macrobase/analysis/stats/mixture/VariationalDPMG.java
```diff
@@ -0,0 +1,288 @@
...
+                break;
+            } else {
+                log.debug("improvement is : {}%", improvement * 100);
+            }
+            log.debug(".........................................");
+        }
+    }
+
+    private void instanciateVariationalParameters(int numPoints, int maxNumClusters) {
```
pbailis: typo: instantiate

src/main/java/macrobase/analysis/transform/BatchScoreFeatureTransform.java
```diff
@@ -28,8 +32,28 @@ public BatchScoreFeatureTransform(MacroBaseConf conf, Iterator<Datum> input) thr
...
+        boolean dumpScores = false;
+        String scoreFile = conf.getString(MacroBaseConf.SCORE_DUMP_FILE_CONFIG_PARAM, "");
+        if (scoreFile != "") {
+            metricsAndDensities = new ArrayList<>(data.size());
+            dumpScores = true;
+        }
         for(Datum d : data) {
+            double score = batchTrainScore.score(d);
             results.add(new Datum(d, batchTrainScore.score(d)));
+            if (dumpScores) {
```
pbailis: @edgan8 is this what you were thinking of doing?

src/main/java/macrobase/analysis/transform/BatchScoreFeatureTransform.java
```diff
@@ -28,8 +32,28 @@ public BatchScoreFeatureTransform(MacroBaseConf conf, Iterator<Datum> input) thr
...
+                metricsAndDensities.add(new MetricsAndDensity(d.getMetrics(), score));
+            }
+        }
+        if (dumpScores) {
```
pbailis: Is it possible to do this inside a different utility class? @edgan8 had some ideas.

src/main/java/macrobase/diagnostics/ScoreDumper.java
pbailis: @edgan8 talked about having a scoredump operator that you could stick into the pipeline. what do you think about that?

pbailis:
Coverage decrease is not phenomenal... :)

pbailis:
LGTM, with a few concerns to address beforehand:

 - Not sure why so many of the log entries need to be at `info`.
 - Would be good to discuss how to dump the scores. @edgan8 had some thoughts.
 - If Travis is failing spuriously, can we use smaller datasets for the tests?

mamikonyana:
>> can we change the coverage check to be slightly more tolerable?

>What do you mean?

Make coverage check not fail if coverage decreased by less than say 0.1% :)

mamikonyana:
> - If Travis is failing spuriously, can we use smaller datasets for the tests?

Travis does not fail any longer after I introduced a limit on number of iterations tests run

>  - Not sure why so many of the log entries need to be at info.

Changed info's into debugs

>  - Would be good to discuss how to dump the scores. @edgan8 had some thoughts.

I reverted `src/main/java/macrobase/analysis/transform/BatchScoreFeatureTransform.java`, I though I had done that before too. I'll talk to @edgan8 about his ideas.

pbailis:
> Make coverage check not fail if coverage decreased by less than say 0.1% :)

1000 * 0.1% = 100%!

src/main/java/macrobase/diagnostics/ScoreDumper.java
viveksjain: Sorry I didn't get to this earlier, but ScoreDumpDiagnostic can be run as a pipeline, right? See `bin/diagnostic.sh`.

PR #109, opened 2016-04-19
Benchmarking scripts

mamikonyana:


pbailis:
LGTM; would be nice to have some docs on how to run this in the wiki.

mamikonyana:
sounds good. I expect the interface to change after we have a mainstream way of dumping densities out of macrobase, so I'll put something in the wiki when that happens.

PR #111, opened 2016-04-22
Allow using DensityEstimators as Classifiers (instead of Transforms)

mamikonyana:
Add some constructs to Macrobase to be able to use Density Estimators (e.g. KDE, GMM, etc.) as bases for classification (instead of using percentile classifier after running those classes as transform)

### DIFF
--- a/src/main/java/macrobase/analysis/stats/mixture/VariationalGMM.java
+++ b/src/main/java/macrobase/analysis/stats/mixture/VariationalGMM.java
@@ -1,6 +1,7 @@
 package macrobase.analysis.stats.mixture;
 
 import macrobase.analysis.stats.BatchMixtureModel;
+import macrobase.analysis.stats.DensityEstimater;
 import macrobase.analysis.stats.distribution.MultivariateTDistribution;
 import macrobase.conf.MacroBaseConf;
 import macrobase.conf.MacroBaseDefaults;
@@ -17,7 +18,7 @@ import java.util.List;
 /**
  * Fit Gaussian Mixture models using Variational Bayes
  */
-public class VariationalGMM extends BatchMixtureModel {
+public class VariationalGMM extends BatchMixtureModel implements DensityEstimater {
     private static final Logger log = LoggerFactory.getLogger(VariationalGMM.class);
 
     private int K;  // Number of mixture components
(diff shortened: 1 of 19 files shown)


mamikonyana:
@edgan8 @pbailis please have a look

@edgan8  I implemented it slightly differently since I need to dump densities on a scored grid also, so I have to generate points and score those and I need access to density estimator. Feel free to suggest an alternative if you think other design would be better.

I also included a script to plot the data that is being dumped.

src/main/java/macrobase/analysis/stats/mixture/VariationalGMM.java
```diff
@@ -17,7 +18,7 @@
 /**
  * Fit Gaussian Mixture models using Variational Bayes
  */
-public class VariationalGMM extends BatchMixtureModel {
+public class VariationalGMM extends BatchMixtureModel implements DensityEstimater {
```
pbailis: Estimator

src/main/java/macrobase/analysis/transform/BatchScoreFeatureTransform.java
```diff
@@ -12,20 +12,25 @@
 public class BatchScoreFeatureTransform extends BatchTransform {
     protected BatchTrainScore batchTrainScore;
     protected MacroBaseConf conf;
+    protected boolean identityTransform;
 
     public BatchScoreFeatureTransform(MacroBaseConf conf, Iterator<Datum> input, MacroBaseConf.TransformType transformType)
             throws ConfigurationException {
         super(input);
-        this.batchTrainScore = conf.constructTransform(transformType);
+        if (transformType == MacroBaseConf.TransformType.IDENTITY) {
```
pbailis: Why is this necessary???

src/main/java/macrobase/conf/MacroBaseConf.java
```diff
@@ -299,6 +345,20 @@ public Boolean getBoolean(String key, Boolean defaultValue) {
...
+            return MacroBaseDefaults.CLASSIFIER_TYPE;
+        }
+        return ClassifierType.valueOf(_conf.get(CLASSIFIER_TYPE));
+    }
+
+    public DensityEstimaterType getDensityEstimatorType() {
+        if (!_conf.containsKey(DENSITY_ESTIMATER_TYPE)) {
+            return MacroBaseDefaults.DENSITY_ESTIMATER_TYPE;
+        }
+        return DensityEstimaterType.valueOf(_conf.get(DENSITY_ESTIMATER_TYPE));
```
pbailis: Estimator, ESTIMATOR

edgan8:
Hmm, maybe I am missing something. It looks like you are still using percentile based cutoffs on the density scores, so what do we gain by having a distinct density classifier? It seems nicer to always use the same common percentile classifier step.

pbailis:
+1; I strongly believe we should have only `FeatureTransform` and `OutlierClassifier` top level classes to handle this functionality.

pbailis:
I can see some scenarios where combining the density estimation and the percentile classification would be beneficial, but I don't see any of them in this PR.

PR #114, opened 2016-04-26
TreeKDE bugfix, consistent data ordering, classifier dumping

edgan8:
Preparing some groundwork for running experiments on tree base KDE. There are three main changes here:

1) Added a new type of classifier, DumpClassifier.

It just wraps another classifier, forwards its output, but also dumps the data to a file. Dumping can be enabled or disabled in the standard pipeline with a conf flag. The dump consists of rows of the form datum #, isOutlier.

2) In order for datum # to make sense I made the data stream ordering immutable. 

Previously the MAD and PercentileClassifier both modified the stream ordering by sorting it. I replaced some usages of sorting with the apache Percentile library but this changes a key behavior: *you are now no longer guaranteed x%* if x is y our percentile if too many of your points have the exact same value. I think this is better than arbitrarily picking some points as outliers.

3) Cleaned up some KDETree logic so that it works well on the energy data:

./bin/batch.sh conf/local_kde.yaml should identify about 6000 outliers.

This required changing to unnormalized densities to avoid underflow, and I also switched to working with log densities for interpretability. The leaf approximation did not work for the energy data so I disabled it while I work on a more refined version.


src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
```diff
@@ -26,32 +34,37 @@ public BatchingPercentileClassifier(MacroBaseConf conf, Iterator<Datum> input) {
...
+        }
+        double[] scores = new double[data.size()];
+        for (int i = 0; i < data.size(); i++) {
+            scores[i] = data.get(i).getNorm();
+        }
+
+        Percentile pCalc = new Percentile().withNaNStrategy(NaNStrategy.MAXIMAL);
+        pCalc.setData(scores);
+        this.cutoff = pCalc.evaluate(scores, targetPercentile * 100);
+        log.info("{} Percentile Cutoff: {}", targetPercentile, cutoff);
```
pbailis: Does this need to be at `info`?

src/main/java/macrobase/analysis/classify/OutlierClassifierImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import macrobase.datamodel.Datum;
+
+import java.util.Iterator;
+
+/**
+ * Consumes a tuple at-a-time from a FeatureTransform and for each tuple outputs
+ * a OutlierClassificationResult that should indicate if the tuple observed was an
+ * outlier or not.
+ */
+public abstract class OutlierClassifierImpl implements OutlierClassifier {
```
pbailis: Why do we need this in addition to `OutlierClassifier`?

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
```diff
@@ -57,15 +59,22 @@ public AnalysisResult contextualAnalyze(DataIngester ingester) throws SQLExcepti
     AnalysisResult run() throws SQLException, IOException, ConfigurationException {
         long startMs = System.currentTimeMillis();
         DataIngester ingester = conf.constructIngester();
+        // TODO: this should be a new pipeline
+        // Needs to happen early before ingester is possibly consumed. Downstream stages are allowed to drain
+        // upstream stages in construction.
+        if(contextualEnabled){
```
pbailis: Agreed // makes sense.

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
```diff
@@ -57,15 +59,22 @@ public AnalysisResult contextualAnalyze(DataIngester ingester) throws SQLExcepti
...
+            return contextualAnalyze(ingester);
+        }
+
         TimedBatchIngest batchIngest = new TimedBatchIngest(ingester);
         FeatureTransform featureTransform = new BatchScoreFeatureTransform(conf, batchIngest, conf.getTransformType());
         OutlierClassifier outlierClassifier = new BatchingPercentileClassifier(conf, featureTransform);
+        if (conf.getBoolean(MacroBaseConf.CLASSIFIER_DUMP, MacroBaseDefaults.CLASSIFIER_DUMP)) {
```
pbailis: Ha, not my favorite hack, but a reasonable compromise. @mamikonyana I think this addresses your concern too, right?

Any chance we can do this in the `MacroBaseConf` object instead?

src/main/java/macrobase/datamodel/KDTree.java
```diff
@@ -60,9 +60,12 @@ public KDTree(List<Datum> data, int leafCapacity) {
             Collections.sort(data, new DatumComparator(splitDimension));
 
             int splitIndex = data.size() / 2;
-            Datum belowSplit = data.get(splitIndex);
-            Datum aboveSplit = data.get(1 + splitIndex);
-            this.splitValue = 0.5 * (aboveSplit.getMetrics().getEntry(splitDimension) + belowSplit.getMetrics().getEntry(splitDimension));
+            Datum belowSplit = data.get(splitIndex - 1);
+            Datum aboveSplit = data.get(splitIndex);
+            this.splitValue = 0.5 * (
+                    aboveSplit.getMetrics().getEntry(splitDimension) +
```
pbailis: spacing?

pom.xml
```diff
@@ -185,5 +185,10 @@
       <artifactId>JRI</artifactId>
       <version>0.9-7</version>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-all</artifactId>
```
pbailis: Just to check, what's this for?

pbailis:
Generally LGTM, some comments inline to address / note.

Also, do you think it makes more sense in the long term to have a separate dumper pipeline?

I am thinking of having a `bin/custom.sh` and corresponding Dropwizard command to allow us to run custom, pre-defined, non-default pipelines. We could even make `batch` and `streaming` act as instances of this. We could even use reflection to load the specific pipeline by name...?

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
edgan8: I thought knowing the rough distribution of percentiles was useful for almost any run, but maybe it's mostly useful for debugging runs. I'm fine with either, what's been your heuristic so far for deciding?

src/main/java/macrobase/analysis/classify/OutlierClassifierImpl.java
edgan8: To share the constructor (the DumperClassifier doesn't use that constructor). Do you think we should just have the constructor explicitly in each classifier since it's very short?

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
edgan8: I agree it's a hack, I'll open switch to a new pipeline ASAP when we get a chance to support running different pipelines with the same command but didn't want to make this PR too big

pom.xml
edgan8: it supports more sophisticated junit assertions like greaterThan, it leads to much more readable test failures than the failure of some opaque boolean flag.

edgan8:
Totally agreed polluting the basic pipeline is bad, I'll move all the dumping logic to a new pipeline when we make it possible to run different pipelines from the command line easily.

src/main/java/macrobase/analysis/classify/BatchingPercentileClassifier.java
pbailis: I leave most internal statistics-related diagnostics at `debug`.

src/main/java/macrobase/analysis/classify/OutlierClassifierImpl.java
pbailis: I'm torn but leaning towards having each class implement its own constructor.

pom.xml
pbailis: Add `<scope>test</scope>`?

pbailis:
LGMT, minor comments. Can you fix and then rebase commits into one?

For merging the PR, rebase master to your branch locally, then push to origin...
https://github.com/stanford-futuredata/macrobase/wiki/Development-Notes#git-workflow

src/main/java/macrobase/analysis/stats/TreeKDE.java
```diff
@@ -79,9 +81,21 @@ public KDTree getKdtree() {
...
+            log.debug("Scored {}", numScored);
+        }
         double unscaledScore = scoreKDTree(kdtree, datum);
         // Note: return score with a minus sign, s.t. outliers are selected not inliers.
-        return -unscaledScore * scoreScalingFactor;
+        return -(Math.log(unscaledScore) + scoreScaleLog);
+    }
+
+    public double scoreDensity(Datum datum) {
+        return -Math.exp(-score(datum));
```
mamikonyana: why is there a minus sign before the exponent?

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
mamikonyana: not really. this class only outputs the result of the classification, which I don't care about.
I care about dumping metrics on datum before the transformation and after the transformation together. There is no place in the pipeline that we currently have access to both.
I guess now I can dump both transforms and classification results, and then combine, but this PR is not letting me do that, I still have to write custom code to accomplish that

src/main/java/macrobase/analysis/stats/TreeKDE.java
edgan8: So that it's still a "score" and outliers have high scores (and to make it consistent with the score KDE returns). I'm refactoring the KDE and TreeKDE class now and I can change the function to just density(Datum datum) though and just return the actual true density since we don't really need a negative density anywhere.

mamikonyana:
> This required changing to unnormalized densities to avoid underflow, and I also switched to working with log densities for interpretability. The leaf approximation did not work for the energy data so I disabled it while I work on a more refined version.

I didn't notice you disabling leaf approximation, if I missed it could you make that a flag? I'm currently using TreeKDE too, I'd really like not to see change in behavior.

pbailis:
If you're depending on a given behavior, unit tests should prevent check-ins that break the behavior.

edgan8:
Though it fixes very poor accuracy on my dataset, disabling the leaf approximation mainly makes TreeKDE slower which is understandably annoying for arsen. I talked with Arsen offline and I will fork the KDE classes for separate experiments in my next diff so I don't block him on his experiments and we can agree on a unified final KDE class some time down the road.

PR #116, opened 2016-04-28
Arbitrary pipelines from command line.

pbailis:
This PR allows users to run arbitrary pipelines from the command line.

Major changes:
 * Removes the `batch` and `streaming` commands from the MacroBase Dropwizard application.
 * Adds a new `pipeline` command to the MacroBase Dropwizard application.
 * `pipeline` dynamically constructs a `Pipeline` instance with class determined by the `macrobase.pipeline.class` configuration parameter.
 * Each `Pipeline` must implement a no-arg constructor and the `initialize(MacroBaseConf)` function.

To use, you can specify the pipeline class in yaml (e.g., `conf/batch.yaml`):
`macrobase.pipeline.class: macrobase.analysis.pipeline.BasicBatchedPipeline`

You can also specify this using command line flags:
`-Dmacrobase.pipeline.class=my.fancy.new.pipeline`
(e.g., `JAVA_OPTS=$JAVA_OPTS" -Dmacrobase.pipeline.class=my.fancy.new.pipeline"; bin/batch.sh`)


bin/batch.sh
```diff
@@ -3,4 +3,4 @@ conf_file=${1:-"conf/batch.yaml"}
 
 set -x
 
-java ${JAVA_OPTS} -cp "src/main/resources/:target/classes:target/lib/*:target/dependency/*" macrobase.MacroBase batch $conf_file
+java ${JAVA_OPTS} -cp "src/main/resources/:target/classes:target/lib/*:target/dependency/*" macrobase.MacroBase pipeline $conf_file
```
mamikonyana: should we just merge `bin/batch.sh` and `bin/streaming.sh`? Or we could force them to be batch and streaming using command line arguments, currently they are pretty much identical.

src/main/java/macrobase/analysis/pipeline/Pipeline.java
```diff
@@ -0,0 +1,9 @@
+package macrobase.analysis.pipeline;
+
+import macrobase.analysis.result.AnalysisResult;
+import macrobase.conf.MacroBaseConf;
+
+public interface Pipeline {
```
mamikonyana: could we have comments on interface methods what their purpose is?

mamikonyana:
nit about putting comments on interfaces and shell scripts. lgtm otherwise

src/main/java/macrobase/runtime/command/MacroBasePipelineCommand.java
```diff
@@ -1,28 +1,37 @@
...
     }
 
     @Override
     protected void run(Bootstrap<MacroBaseConf> bootstrap,
                        Namespace namespace,
                        MacroBaseConf configuration) throws Exception {
         configuration.loadSystemProperties();
-        BasicBatchedPipeline analyzer = new BasicBatchedPipeline(configuration);
+        Class c = Class.forName(configuration.getString(MacroBaseConf.PIPELINE_NAME));
+        Object ao = c.newInstance();
```
edgan8: If we don't want to do this initialize stuff, does using getConstructor as described here work? http://stackoverflow.com/a/6094600

src/main/java/macrobase/analysis/pipeline/Pipeline.java
```diff
@@ -0,0 +1,9 @@
...
+    void initialize(MacroBaseConf conf) throws Exception;
```
edgan8: If we have this return a Pipeline we can do stuff like (new Pipeline()).initialize(x).run() on a single line which is nice.

edgan8:
Thanks, this will be great for testing! LGTM besides the one ease-of-coding suggestion.

bin/batch.sh
pbailis: How do you propose merging?

src/main/java/macrobase/analysis/pipeline/Pipeline.java
pbailis: Sure, what were you thinking here? Happy to add...

src/main/java/macrobase/analysis/pipeline/Pipeline.java
pbailis: Will add in the next fluent PR to avoid merge issues.

pbailis:
Thanks, @mamikonyana @edgan8!

bin/batch.sh
mamikonyana: config file could be the first argument to the script, like it is now with `bin/batch.sh`. Otherwise they are identical, no?

src/main/java/macrobase/analysis/pipeline/Pipeline.java
mamikonyana: A once sentence description what each method is for. It took me some time to understand the new iterator-like interfaces so having something as a documentation would be helpful.

PR #117, opened 2016-04-28
GridDumpingPipeline

mamikonyana:
Creates a new pipeline that will dump scores both into a density grid and before/after score pairs of batchFeatureTransform

main goal of this PR is to be able to dump scored grid of the data

### DIFF
--- a/src/main/java/macrobase/util/DiagnosticsUtils.java
+++ b/src/main/java/macrobase/util/DiagnosticsUtils.java
@@ -9,18 +9,13 @@ import java.util.List;
 
 public class DiagnosticsUtils {
 
-    public static List<Datum> create2DGrid(double[][] boundaries, double delta) {
+    public static List<Datum> createGridFixedIncrement(double[][] boundaries, double delta) {
         int dimension = boundaries.length;
         int[] dimensionPoints = new int[dimension];
         int size = 1;
         for (int d = 0; d < dimension; d++) {
             dimensionPoints[d] = (int) ((boundaries[d][1] - boundaries[d][0]) / delta + 1);
             size *= dimensionPoints[d];
-        }
-        List<Datum> scoredData = new ArrayList<>((int) size);
-        double[] point = new double[dimension];
-        for (int d = 0; d < dimension; d++) {
-            point[d] = boundaries[d][0];
         }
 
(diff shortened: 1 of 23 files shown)


mamikonyana:
@pbailis @edgan8 could you take a look?

src/main/java/macrobase/runtime/standalone/MacroBaseGridCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import io.dropwizard.setup.Bootstrap;
+import macrobase.MacroBase;
+import macrobase.analysis.pipeline.GridDumpingPipeline;
+import macrobase.analysis.result.AnalysisResult;
+import macrobase.conf.MacroBaseConf;
+import net.sourceforge.argparse4j.inf.Namespace;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MacroBaseGridCommand extends ConfiguredCommand<MacroBaseConf> {
```
pbailis: This won't be necessary after the command line refactor, but okay for now.

src/main/java/macrobase/util/DiagnosticsUtils.java
```diff
@@ -9,19 +9,14 @@
 
 public class DiagnosticsUtils {
 
-    public static List<Datum> create2DGrid(double[][] boundaries, double delta) {
+    public static List<Datum> createGridFixedIncrement(double[][] boundaries, double delta) {
```
pbailis: Question: why are we dumping a grid instead of just dumping all scores, then post-processing?

pbailis:
LGTM; some of the code won't be necessary after the command line push. You can either remove after I push, or I can remove after you push.

Before merging, one question: why compute the grid inside MB instead of just dumping scores and computing it later?

src/main/java/macrobase/util/DiagnosticsUtils.java
mamikonyana: I want to actually see the distribution in areas without data, also it makes plotting much easier.

mamikonyana:
I'll rebase off master and remove unnecessary code before merging (maybe later tonight)

PR #118, opened 2016-04-28
Fluent Pipeline API

pbailis:
This PR provides a nicer way to write chains of operators.

For example:

        MBStream<Summary> s = OperatorChain.begin(data)
                                           .then(new BatchScoreFeatureTransform(conf, conf.getTransformType()))
                                           .then(new BatchingPercentileClassifier(conf))
                                           .then(new BatchSummarizer(conf))
                                           .execute();

There is also a microBatch API, which polls a configurable number of tuples from each operator in the chain (same constructor pattern with `then`, but different `execute` methods that accept the maximum number of tuples per batch passed between each operator).

You can see an example of microbatch [in the streaming pipeline](https://github.com/stanford-futuredata/macrobase/blob/e377fd9ed8efba3bf3e3782a8a867cf5d1e700de/src/main/java/macrobase/analysis/pipeline/BasicOneShotEWStreamingPipeline.java#L36): we execute the feature transform and classification incrementally, 1000 tuples at a time, until we run out of tuples to process (i.e., we hit fixpoint). Then we pass the entire stream to the summary stage.

The downside of this API is that it's pushing the limits of Java's type inference and sometimes requires love in terms of casting things. We also had to create a `OneToOneOperator` class to get around the fact that parameters can only implement a single interface.

The plus side of this API is that it makes the code marginally more readable (IMO) and is also type safe.

Happy to iterate, but figured it might be helpful (and an interesting experiment in how we can use the `MBStream` interface to play with things like micro-batching).

### DIFF
--- a/src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
+++ b/src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
@@ -5,22 +5,18 @@ import macrobase.analysis.classify.DumpClassifier;
 import macrobase.analysis.classify.OutlierClassifier;
 import macrobase.analysis.contextualoutlier.Context;
 import macrobase.analysis.contextualoutlier.ContextualOutlierDetector;
+import macrobase.analysis.pipeline.operator.MBOperatorChain;
 import macrobase.analysis.result.AnalysisResult;
 import macrobase.analysis.stats.BatchTrainScore;
 import macrobase.analysis.summary.BatchSummarizer;
 import macrobase.analysis.summary.Summary;
-import macrobase.analysis.transform.FeatureTransform;
-import macrobase.conf.ConfigurationException;
 import macrobase.conf.MacroBaseConf;
-import macrobase.conf.MacroBaseDefaults;
 import macrobase.datamodel.Datum;
 import macrobase.ingest.DataIngester;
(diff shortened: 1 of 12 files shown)


pbailis:
If this goes in, it should go in after #116.

@mamikonyana @edgan8

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
```diff
@@ -72,24 +68,24 @@ public AnalysisResult run() throws Exception {
             return contextualAnalyze();
         }
 
+
```
mamikonyana: empty space

mamikonyana:
lgtm

src/main/java/macrobase/analysis/pipeline/operator/OneToOneOperator.java
```diff
@@ -0,0 +1,3 @@
+package macrobase.analysis.pipeline.operator;
+
+public interface OneToOneOperator<S, T> extends MBConsumer<S>, MBProducer<T> { }
```
edgan8: What do you think about renaming this to just Operator? When we have multi operators we can specify that in the name

pbailis: Will do.

PR #124, opened 2016-05-03
Refactor of fluent operator to use Operator instead of OperatorChain.

pbailis:
With @edgan8 and @sahaana!


src/main/java/macrobase/analysis/pipeline/operator/MBOperator.java
```diff
@@ -1,3 +1,33 @@
...
+    public <Y> MBOperator<S, Y> then(MBOperator<T, Y> o2, int batchSize) {
+        MBOperator<S, T> self = this;
+        return new MBOperator<S, Y>() {
+            @Override
+            public MBStream<Y> getStream() throws Exception {
+                return o2.getStream();
+            }
+
+            @Override
+            public void initialize() throws Exception { }
```
edgan8: If we want to actually use initialize in the future we probably want this chained operator to be like

```
public void initialize() throws Exception{
  self.initialize();
  o2.initialize();
}
```

and similarly with shutdown

src/main/java/macrobase/analysis/pipeline/stream/MBMultiInputStream.java
```diff
@@ -0,0 +1,41 @@
...
+    public MBMultiInputStream() {
+        this(new ArrayList<>());
+    }
+
+    public void addStream(MBStream<T> input) {
+        inputs.add(input);
+    }
+
+    @Override
+    public List<T> drain(int maxElements) {
```
edgan8: Can we add a test for streaming usage of this class? Getting streaming mini-batch draining to work seems tricky.

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
```diff
@@ -42,16 +43,14 @@ public Pipeline initialize(MacroBaseConf conf) throws Exception {
             outlierClassifier = new DumpClassifier(conf, outlierClassifier, queryName);
         }
 
-        MBOperatorChain<?, Summary> chain = MBOperatorChain.begin(data)
-                .then(new BatchScoreFeatureTransform(conf, conf.getTransformType()))
+        MBOperator<Datum, Summary> op =
```
edgan8: In cases like the main basic batched pipeline, what do you think about working with unwrapped operators and passing around their output directly? This will probably be easier to debug in the stack trace and dump debugging output from, while still being almost the same number of lines of code. 

One big advantage of the chaining syntax is that it allows us to wrap up multiple operators into one when we need it e.g. in the groupedby operator, but otherwise I thought the intermediate syntax of getStream.drain() was fine.

src/main/java/macrobase/analysis/pipeline/operator/MBOperator.java
pbailis: Good catch, thanks!

src/main/java/macrobase/analysis/pipeline/stream/MBMultiInputStream.java
pbailis: Will do.

src/main/java/macrobase/analysis/pipeline/BasicBatchedPipeline.java
pbailis: Fine by me.

src/main/java/macrobase/analysis/pipeline/stream/MBMultiInputStream.java
pbailis: Added: 53d126b

pbailis:
Comments addressed.

PR #106, opened 2016-04-18
rebase contextual outlier detection to refactored code 

raininthesun:
Major changes include:
-- rebase contextual outlier detection to refactored code
-- add another API to contextual outlier detection that takes as input some outliers expressed as predicates. documentation is also added to `parameters_desc.md`
-- Change Math package <artifactId>commons-math3</artifactId> to version 3.6 in pom.xml (do mvn clean; mvn package)
-- add BasicContextualBatchedPipeline. (invoke the program using `java ${JAVA_OPTS} -cp "src/main/resources/:target/classes:target/lib/*:target/dependency/*" macrobase.MacroBase contextualbatch  batch.yaml`
-- change DiskCachingIngester to include contextual attributes

PR #105, opened 2016-04-11
added histograms to frontend

vayu-stanford:
Histograms mainly added through .js and .html files. New
MultipleRowSetResource class added to facilitate querying of multiple
datasets from server

Note that this commit introduces a new dependency on plotly.js, which is
MIT licensed

PR #128, opened 2016-05-05
Bug fix: frontend should expect list of results.

pbailis:
Following #106, pipelines return a list of AnalysisResult. The frontend expects a single AnalysisResult. There are at least two fixes:

 1. Have the frontend render the first AnalysisResult in the returned JSON.
 2. Have the server return the first AnalysisResult from the pipeline.

Option 1 seems better in the event that we ever want to do something with multiple AnalysisResult instances.

@vayu-stanford: This addresses the frontend bug.

It'd be nice to have end-to-end unit tests for the frontend, but, for now, I think it's probably overkill. This is the first instance I'm aware of that actually broke the frontend.


vayu-stanford:
This looks good to me. I will wait till this is merged into master and update PR #105 with the change to avoid any merge conflicts.

PR #129, opened 2016-05-05
Basic sliding window

kexinrong:
- Implemented naive window aggregate function classes: BatchedWindowMax, BatchedWindowCount, BatchedWindowSum
- Implemented simple sliding window transform that outputs one datum for each pane.

### DIFF
--- a/src/main/java/macrobase/analysis/transform/SlidingWindowTransform.java
+++ b/src/main/java/macrobase/analysis/transform/SlidingWindowTransform.java
@@ -7,45 +7,155 @@ import macrobase.conf.ConfigurationException;
 
 import macrobase.conf.MacroBaseDefaults;
 import macrobase.datamodel.Datum;
+import org.apache.commons.math3.linear.ArrayRealVector;
 
 import java.util.ArrayList;
 import java.util.List;
 
 
 public class SlidingWindowTransform extends FeatureTransform {
-    public int minPaneSize;
-    public BatchedWindowAggregate windowAggregate;
-    public int timeColumn;
+    private final BatchedWindowAggregate windowAggregate;
+    private final int timeColumn;
+    private final boolean paneEnabled;
+    private final int slideSize;
+    private int windowSize;
+    private int paneSize;
+    private int windowStart;
+    private int paneStart;
+    private int dim = 0;
 
-    private final MBStream<Datum> output = new MBStream<>();
+    private MBStream<Datum> output = new MBStream<>();
+    private List<Datum> newSlide = new ArrayList<>();
+    private List<Datum> currPane = new ArrayList<>();
+    private List<Datum> currWindow = new ArrayList<>();
 
-    public SlidingWindowTransform(MacroBaseConf conf, int minPaneSize)
+    private int GCD(int a, int b) { return b==0 ? a : GCD(b, a % b); }
+
+    public SlidingWindowTransform(MacroBaseConf conf, int windowSize)
             throws ConfigurationException  {
         MacroBaseConf.AggregateType aggregateType = conf.getAggregateType();
         this.windowAggregate = conf.constructAggregate(aggregateType);
         this.timeColumn = conf.getInt(MacroBaseConf.TIME_COLUMN, MacroBaseDefaults.TIME_COLUMN);
-        this.minPaneSize = minPaneSize;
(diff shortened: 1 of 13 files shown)


src/main/java/macrobase/analysis/transform/SlidingWindowTransform.java
```diff
@@ -0,0 +1,67 @@
...
+import macrobase.datamodel.Datum;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class SlidingWindowTransform extends FeatureTransform {
+    public int minPaneSize;
+    public BatchedWindowAggregate windowAggregate;
+    public int timeColumn;
```
pbailis: Is there any reason these are public? I would expect `private final`

src/main/java/macrobase/analysis/transform/aggregate/BatchedWindowAggregate.java
```diff
@@ -0,0 +1,12 @@
+package macrobase.analysis.transform.aggregate;
+
+import macrobase.datamodel.Datum;
+import java.util.List;
+
+public abstract class BatchedWindowAggregate implements IncrementalWindowAggregate {
```
pbailis: What if instead this class implements `process(List<Datum> new_data, List<Datum> expired_data)` (from `IncrementalWindowAggregate`?

Basic idea: BatchedWindowAggregate can keep track of the current window, then, each time `process(new_data, expired_data)` is called, it updates the window and passes the entire window to an abstract `public Datum process(List<Datum> data)` function.

src/main/java/macrobase/analysis/transform/SlidingWindowTransform.java
```diff
@@ -0,0 +1,67 @@
...
+import macrobase.conf.MacroBaseDefaults;
+import macrobase.datamodel.Datum;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class SlidingWindowTransform extends FeatureTransform {
+    public int minPaneSize;
+    public BatchedWindowAggregate windowAggregate;
```
pbailis: Can this instead be `IncrementalWindowAggregate`?

src/main/java/macrobase/analysis/transform/aggregate/IncrementalWindowAggregate.java
```diff
@@ -0,0 +1,8 @@
+package macrobase.analysis.transform.aggregate;
+
+import macrobase.datamodel.Datum;
+import java.util.List;
+
+public interface IncrementalWindowAggregate {
```
pbailis: Or just `WindowedAggregate`!

src/main/java/macrobase/analysis/transform/SlidingWindowTransform.java
pbailis: Put another way, if every aggregate implements `IncrementalWindowAggregate`, then all this function needs to do is keep track of panes and when they enter and expire...

src/test/java/macrobase/analysis/transform/SlidingWindowTransformTest.java
```diff
@@ -0,0 +1,69 @@
...
+            Datum d = transformed.get(i);
+            assertTrue(d.getMetrics().getEntry(0) == i * 3);
+            if (i == 3)
+                assertTrue(d.getMetrics().getEntry(1) == 1);
+            else
+                assertTrue(d.getMetrics().getEntry(1) == 3);
+
+        }
+
+        /* Test a different aggregation function */
```
pbailis: Minor: would be nice to have these in separate functions to help tell what part of the functionality being tested fails (if and when it fails)!

kexinrong:
Main changes include:
- Change the WindowAggregate interface from _process(List<Datum> data)_ to _updateWindow(List<Datum> new_data, List<Datum> old_data)_
- Each window aggregate class can indicate whether they support panes. If so, the _updateWindow_ function takes in lists of panes; otherwise it takes in lists of datum.
- Refactored _SlidingWindowTransform_ so that it produces a stream of window aggregates. Incomplete window at the end of the data stream can be flushed out during shutdown. 

This ended up being more complicated than I thought. Happy to discuss the changes online/in person.

pbailis:
@kexinrong LMK when you have an update to look at!

kexinrong:
Please refer to PR #136.

PR #132, opened 2016-05-06
Use log density as score for Mixture models

mamikonyana:
This allows to run them with percentileClassifier, because percentile classifier needs high values for outliers not small.


mamikonyana:

mamikonyana:
java changes are in #134, this are changes to accompanying scripts to run inference and plot, not necessarily only for those changes.

pbailis:
@mamikonyana why not combine?

pbailis:
LGTM, would prefer to squash with existing PR, though.

mamikonyana:
I was using this scripts in other branch and I really didn't want to have merge conflicts when I squashed the other PR.

PR #136, opened 2016-05-12
Implement basic sliding window transform for Datum

kexinrong:
Refactor for PR #129.

- Adding interfaces for batch, incremental and pane aggregate
- Implemented window aggregate functions: MAX, SUM, COUNT
- Implemented incremental and batch sliding window transform

### DIFF
--- a/src/main/java/macrobase/analysis/transform/BatchSWTransform.java
+++ b/src/main/java/macrobase/analysis/transform/BatchSWTransform.java
@@ -1,0 +1,60 @@
+package macrobase.analysis.transform;
+
+import macrobase.analysis.pipeline.stream.MBStream;
+import macrobase.analysis.transform.aggregate.BatchWindowAggregate;
+import macrobase.conf.ConfigurationException;
+import macrobase.conf.MacroBaseConf;
+import macrobase.conf.MacroBaseDefaults;
+import macrobase.datamodel.Datum;
+
+import java.util.List;
+
+public class BatchSWTransform extends SlidingWindowTransform {
+    private BatchWindowAggregate windowAggregate;
+
+    public BatchSWTransform(MacroBaseConf conf, int slideSize) throws ConfigurationException {
+        MacroBaseConf.AggregateType aggregateType = conf.getAggregateType();
+        this.windowAggregate = conf.constructBatchAggregate(aggregateType);
+        this.timeColumn = conf.getInt(MacroBaseConf.TIME_COLUMN, MacroBaseDefaults.TIME_COLUMN);
+        this.slideSize = slideSize;
+        this.windowSize = conf.getInt(MacroBaseConf.WINDOW_SIZE, MacroBaseDefaults.WINDOW_SIZE);
+    }
+
+    private void slideWindow() {
+        int i = 0;
+        while (i < currWindow.size() && datumInRange(currWindow.get(i), windowStart, 0)) { i++; }
+        currWindow.subList(0, i).clear();
+    }
+
+    private void aggregateWindow() {
+        slideWindow();
+        Datum newWindow = windowAggregate.aggregate(currWindow);
+        newWindow.getMetrics().setEntry(timeColumn, windowStart);
+        output.add(newWindow);
+        windowStart += this.slideSize;
+    }
+
+    @Override
+    public void consume(List<Datum> data) {
+        if (data.isEmpty())
+            return;
+        if (windowStart < 0)
+            windowStart = data.get(0).getTime(timeColumn);
+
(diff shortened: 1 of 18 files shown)


src/main/java/macrobase/conf/MacroBaseDefaults.java
```diff
@@ -33,6 +33,7 @@
     
     // timeseries defaults
     public static final Integer TUPLE_WINDOW = 100;
+    public static final Integer SLIDE_SIZE = 86400; // in seconds
```
pbailis: Any reason you chose seconds here instead of milliseconds?

src/main/java/macrobase/analysis/transform/BatchSWTransform.java
```diff
@@ -0,0 +1,60 @@
...
+import macrobase.conf.MacroBaseConf;
+import macrobase.conf.MacroBaseDefaults;
+import macrobase.datamodel.Datum;
+
+import java.util.List;
+
+public class BatchSWTransform extends SlidingWindowTransform {
+    private BatchWindowAggregate windowAggregate;
+
+    public BatchSWTransform(MacroBaseConf conf, int windowSize) throws ConfigurationException {
```
pbailis: Any reason windowSize is passed in but not SLIDE_SIZE?

src/main/java/macrobase/analysis/transform/aggregate/IncrementalWindowCount.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class IncrementalWindowCount extends IncrementalWindowAggregate {
+    private int count = 0;
+
+    public IncrementalWindowCount(){}
+
+    public IncrementalWindowCount(MacroBaseConf conf) throws ConfigurationException {
+        this.timeColumn = conf.getInt(MacroBaseConf.TIME_COLUMN, MacroBaseDefaults.TIME_COLUMN);
```
pbailis: Can move this to `IncrementalWindowAggregate` then call `super(conf)` here.

src/main/java/macrobase/analysis/transform/aggregate/IncrementalWindowSum.java
```diff
@@ -0,0 +1,52 @@
...
+import macrobase.conf.MacroBaseDefaults;
+import macrobase.datamodel.Datum;
+import org.apache.commons.math3.linear.ArrayRealVector;
+import org.apache.commons.math3.linear.RealVector;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class IncrementalWindowSum extends IncrementalWindowAggregate {
+    public IncrementalWindowSum(){}
```
pbailis: Is this needed? If not, can remove.

src/main/java/macrobase/analysis/transform/aggregate/IncrementalWindowCount.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.commons.math3.linear.ArrayRealVector;
+import org.apache.commons.math3.linear.RealVector;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class IncrementalWindowCount extends IncrementalWindowAggregate {
+    private int count = 0;
+
+    public IncrementalWindowCount(){}
```
pbailis: Is this needed? If not, can remove.

pbailis:
Very minor comments inline, then LGTM!

src/main/java/macrobase/analysis/transform/BatchSWTransform.java
kexinrong: I thought slide size is fixed. We can pass in slide size as well.

src/main/java/macrobase/conf/MacroBaseDefaults.java
kexinrong: Will change to milliseconds.

src/main/java/macrobase/analysis/transform/BatchSWTransform.java
pbailis: I was thinking the opposite: what about setting window size in conf?

If we want different window sizes for different operators, we can change the conf instance.

The goal of the current conf design is to avoid passing in all params at construction and instead going through conf. This simplifies a few things, including reasoning about default args.

src/main/java/macrobase/conf/MacroBaseDefaults.java
pbailis: :100:

PR #139, opened 2016-05-13
Add Mixture Model parameters to docs

mamikonyana:
@raininthesun hope this helps a little


src/main/java/macrobase/analysis/stats/mixture/FiniteGMM.java
```diff
@@ -38,6 +38,10 @@ public void train(List<Datum> data) {
         VariationalInference.trainMeanField(this, data, mixingComponents, clusters);
     }
 
+    /**
+     * @param datum
+     * @return log probability density of the given datum
+     */
     @Override
```
raininthesun: Is the probability density of the given datum corresponds to the likelihood of the datum? Say  the probability density of the given datum is 0.9, then does it mean that 90% this datum is not an outlier.

mamikonyana: this algorithm does not care how you think about outliers. probability density of the given datum being 0.9 means that there is `0.9 * ds` chance of a datum appearing within `ds` volume of the space surrounding the metrics of the datum.
Basically this number has the same meaning as the one you get from KDE

raininthesun: Got it! Thanks.

raininthesun:
LGTM

PR #140, opened 2016-05-13
Put temporary files in target/ so git tree is clean after `mvn test`

mamikonyana:
@raininthesun could you take a look?


raininthesun:
LGTM.

PR #141, opened 2016-05-13
Tweakable SVI + Accuracy Tasks

mamikonyana:
This PR contributes the following

1. 2 tasks for diagnostics
  - `HeldOutDataLogLikelihoodCalc` which splits data into train/test parts and calculates test data log likelihood after training the model on train data
  - `TrueDensityISECalculator` calculates model density values with true density that was used to generate `src/test/resources/data/3gaussians-7000points.csv.gz`
2. allows user to control Stochastic Variational Inference (see docs/)
3. scripts for plotting convergence when training Variational models

### DIFF
--- a/src/main/java/macrobase/analysis/stats/mixture/BatchMixtureModel.java
+++ b/src/main/java/macrobase/analysis/stats/mixture/BatchMixtureModel.java
@@ -15,20 +15,27 @@ import org.slf4j.LoggerFactory;
 
 import java.io.FileNotFoundException;
 import java.io.FileReader;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
 
 public abstract class BatchMixtureModel extends BatchTrainScore {
     private static final Logger log = LoggerFactory.getLogger(BatchMixtureModel.class);
     protected final double progressCutoff;
     protected final int maxIterationsToConverge;
     protected MacroBaseConf conf;
+    protected double trainTestSplit;
+    protected final String initialClusterCentersFile;
 
     public BatchMixtureModel(MacroBaseConf conf) {
         super(conf);
         this.conf = conf;
         progressCutoff = conf.getDouble(MacroBaseConf.ITERATIVE_PROGRESS_CUTOFF_RATIO, MacroBaseDefaults.ITERATIVE_PROGRESS_CUTOFF_RATIO);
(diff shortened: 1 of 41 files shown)


mamikonyana:

src/main/java/macrobase/analysis/stats/mixture/BatchMixtureModel.java
```diff
@@ -91,7 +100,7 @@ public double getZScoreEquivalent(double zscore) {
     }
 
     public boolean checkTermination(double logLikelihood, double oldLogLikelihood, int iteration) {
-        log.debug("log likelihood after iteration {} is {}", iteration, logLikelihood);
+        log.info("average point log likelihood after iteration {} is {}", iteration, logLikelihood);
```
pbailis: Why info?

src/main/java/macrobase/analysis/stats/mixture/NormalWishartClusters.java
```diff
@@ -106,12 +106,25 @@ public void initializeBaseForDP(List<Datum> data) {
...
-        loc = BatchMixtureModel.gonzalezInitializeMixtureCenters(data, K, random);
+        if (filename != null) {
+            try {
+                loc = BatchMixtureModel.initializeClustersFromFile(filename, K);
+                log.debug("loc : {}", loc);
+                if(loc.size() < K) {
+                    loc = BatchMixtureModel.gonzalezInitializeMixtureCenters(loc, data, K, random);
+                }
+            } catch (FileNotFoundException e) {
+                e.printStackTrace();
```
pbailis: Is this expected behavior?

pbailis:
Very minor comments; looks fine.

src/main/java/macrobase/analysis/stats/mixture/BatchMixtureModel.java
mamikonyana: debug doesn't show up when I'm running diagnostic tasks. I didn't know how to set a not default log level there.

pbailis: You should be able to change the logging level in your MacroBase yaml conf.
For example, see the server conf.

src/main/java/macrobase/analysis/stats/mixture/NormalWishartClusters.java
mamikonyana: if file is provided, initialize K points from file, or if file contains less than K points in the file, take those points as seeds. I'll add this as a comment

PR #142, opened 2016-05-25
Add MixtureGroupClassifier that runs summarization on all the groups …

mamikonyana:
…in a mixture model


mamikonyana:
This PR accomplishes the following:

- Allows MixtureModelPipeline to run MixtureGroupClassifier on all of the clusters if a targetGroup setting is not set.
- Adds python scripts that plot iterative algorithm fitting process
- Adds python scripts that allow showing summaries when clicking on clusters in a matplotlib window.

mamikonyana:

pbailis:
LGTM

PR #145, opened 2016-05-29
QuickSelect MCD and MAD; minor FPG speedup.

pbailis:
Adds quickselect to MAD and MCD.

### DIFF
--- a/src/main/java/macrobase/analysis/stats/MinCovDet.java
+++ b/src/main/java/macrobase/analysis/stats/MinCovDet.java
@@ -24,7 +24,6 @@ import org.apache.commons.math3.linear.RealMatrix;
 import org.apache.commons.math3.linear.RealVector;
 import org.apache.commons.math3.linear.SingularMatrixException;
 import org.apache.commons.math3.linear.SingularValueDecomposition;
-import org.ddogleg.sorting.QuickSelect;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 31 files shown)


pbailis:
I have no idea why https://github.com/stanford-futuredata/macrobase/pull/145#issuecomment-222595941 reduced coverage... It's a one-line fix.

pbailis:
This is ready for review. @sahaana @kexinrong @edgan8 @mamikonyana

src/main/java/macrobase/analysis/stats/MinCovDet.java
```diff
@@ -148,19 +149,34 @@ private RealVector getMean(List<? extends HasMetrics> data) {
...
             return scores;
         }
 
-        scores.sort((a, b) -> a.getScore().compareTo(b.getScore()));
+        double cutoff = QuickSelect.select(scoreVec, k, scoreVec.length);
+
+        List<MetricsWithScore> ret = new ArrayList<>(k);
 
-        return scores.subList(0, k);
+        for(MetricsWithScore mws : scores) {
```
kexinrong: I though QuickSelect.select() sorts the given array up to the kth element so you don't need to traverse through the whole thing again?

pbailis: I think you're right, but QuickSelect.select() operates on the array of doubles, not the `scores` array. I think the best way would be to write a QuickSelect implementation that operates on Java comparables so we could do this and also avoid a copy. I wasn't going to do it this time around, but maybe it's worth it.

kexinrong: I see. I saw in the [javadoc](http://ddogleg.org/javadoc/index.html?org/ddogleg/sorting/QuickSelect.html) that there's a ` public static Comparable select(Comparable[] data, int k, int maxIndex, int[] indexes)` interface, where `indexes` is the (output) sorted list of indexes. Maybe you can use this instead of writing your own implementation?

pbailis: I ended up removing this, since it seems to be slower on some of our datasets. 👍

src/main/java/macrobase/ingest/DiskCachingIngester.java
```diff
@@ -48,42 +50,18 @@ public String getBaseQuery() {
...
-    }
-
-
     private void initialize() throws Exception {
         if(output == null) {
             List<Datum> data;
             data = readInData();
             if (data == null || data.size() == 0) {
-                output = innerIngester.getStream();
+		data = innerIngester.getStream().drain();
```
mamikonyana: fix indent.

pbailis:
@mamikonyana: fixed, thanks! Anything else?

mamikonyana:
lgtm

PR #155, opened 2016-07-18
Allow specifying JDBC properties in a config file as a dictionary

mamikonyana:


pbailis:
Is there a reason why this patch subclasses `PostgreIngester` instead of just adding a new configuration flag?

mamikonyana:
You'd prefer to have an option to pass a dictionary of properties that we'd like to use as JDBC properties?

mamikonyana:
@pbailis

pbailis:
LGTM; pls squash commits before merge.

PR #156, opened 2016-07-22
BigDAWG Ingester

mamikonyana:
Create an Ingester that connects to the BIG DAWG database, works with the web UI.

### DIFF
--- a/bench/conf/workflow_config.json
+++ b/bench/conf/workflow_config.json
@@ -398,5 +398,103 @@
     "macrobase.loader.targetHighMetrics": [
       "dis_amo"
     ]
+  },
+  {
+    "macrobase.loader.targetLowMetrics": [],
+    "macrobase.loader.db.baseQuery": "SELECT * FROM network LIMIT 4000000;",
+    "isBatchJob": true,
+    "macrobase.loader.attributes": [
+      "hostname"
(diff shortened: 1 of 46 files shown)


pbailis:
Since this is a custom ingester, I'd prefer not to put it in the main source tree.

I'm wondering if it's time to create a new Maven submodule (e.g., `tools`) that contains code like this. 

Then, we could support a `DataLoader.CUSTOM` enum, which requires an extra arg `macrobase.loader.class` that can be used to dynamically load the custom loader, as we do with pipelines.

What do you think?

mamikonyana:
I think it's too early to separate custom loaders into a separate submodule. It makes things complicated, but if you really don't want to put this in the main source tree, I could try creating a new submodule, but I'm worried that it will become stale soon after we do that.

pbailis:
@mamikonyana: do you still want to merge this?

pbailis:
(Unintentionally closed, sorry!)

pbailis:
@mamikonyana do you have plans to merge this?

mamikonyana:
I was under impression you didn't want this.. If you want this merged, I can update the revision to work with new repository setup.

pbailis:
Closing for now.

PR #159, opened 2016-08-10
Google Monitoring API ingester

bryankate:
Add a Google Monitoring API ingester, which provides access to monitoring data for resources running in GCP (and AWS if the user is a Stackdriver customer).

This is tested in batch mode using Application Default Credentials to authenticate.

### DIFF
--- a/src/main/java/macrobase/ingest/GoogleMonitoringIngester.java
+++ b/src/main/java/macrobase/ingest/GoogleMonitoringIngester.java
@@ -1,0 +1,340 @@
+package macrobase.ingest;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.util.Lists;
+import com.google.api.services.monitoring.v3.Monitoring;
+import com.google.api.services.monitoring.v3.Monitoring.Projects;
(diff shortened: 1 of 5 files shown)


src/main/java/macrobase/ingest/GoogleMonitoringIngester.java
```diff
@@ -0,0 +1,340 @@
...
+    private int pointsAdded = 0;
+    private int skippedTimeSeries = 0;
+    private int skippedPoints = 0;
+
+    public GoogleMonitoringIngester(MacroBaseConf conf) throws ConfigurationException, IOException {
+        super(conf);
+    }
+
+    @Override
+    public String getBaseQuery() {
```
pbailis: We should get rid of this... Will do so in a later refactor.

src/main/java/macrobase/ingest/GoogleMonitoringIngester.java
```diff
@@ -0,0 +1,340 @@
...
+            }
+        }
+    }
+
+    // Package scope to allow testing.
+    Datum processRecord(Record rec) throws Exception {
+        int idx = 0;
+        RealVector metricVec = new ArrayRealVector(lowMetrics.size() + highMetrics.size());
+        for (String metric : lowMetrics) {
+            metricVec.setEntry(idx, Math.pow(Math.max(rec.values.get(metric), 0.1), -1));
```
pbailis: Another hack we should also get rid of :) Would love to discuss the best way to do so.

pbailis: Fine for now, something I'd like to remove from the project going forwards.

pbailis:
LGTM, thanks, @bryankate! Can also discuss tomorrow. Thanks for slogging through the current `DataIngester` interface.

Issue #147, opened 2016-06-02
Proposal: move `bench` under `src/main/test`


pbailis:
Any thoughts? @mamikonyana @kexinrong @sahaana @edgan8 @raininthesun?

pbailis:
Ditto `scripts`... 

Also, @mamikonyana: how is `scripts/benchmark` different than `scripts/bench`? Should we merge them?

mamikonyana:
We should merge `/scripts/benchmark` and `/bench`, I created the former when I had forgotten that we had scripts under `/bench`.

what's the benefit of having `bench` under `src/main/test`? It seems cleaner having python code in a separate high level directory

pbailis:
Great, will work on consolidation.

`bench` is a type of test, would prefer to keep the root directory clean (e.g., Spark, Mesos, don't have this). Alternatively, could create a `tools/` dir?

@kexinrong @edgan8 @sahaana?

edgan8:
I agree with @mamikonyana that it's nice for the toplevel "test" directory be reserved for relatively fast-running correctness tests. We can then have benchmarks in a separate directory since they're probably not going to be run when you want "mvn test". I've been thinking about creating an "experiments" directory where we can have standalone code we use to try certain hypotheses without polluting the main source or the actual correctness tests, what do you guys think? General purpose "scripts" that are broadly useful to normal macrobase usage (if there are any) can then also have their own directory.

pbailis:
I'm okay with an 'experiments' subdir, though we'll need to modify the Maven layout. What were you thinking: creating a submodule?

pbailis:
Done via #160

PR #160, opened 2016-08-11
Refactor: submodules, loaders

pbailis:
In preparation for increased repo activity, this patch does the following:
  * Migrates to a multi-module format. Core MB operators are in `core`, while new pipelines and features are in `tools` and will migrate to `core` over time. The frontend is in `frontend`.
  * Cleans up the `Ingester` interface. Specifically, we're seeing a lot of activity here, and simplifying ETL will be a big win. We still have `targetLowMetrics` but only three fields for now. Removed `getBaseQuery()`, which is inappropriate for non-SQL loaders.
  * Migrates experimental features to `tools`.

Feedback appreciated. In particular, we could have a separate `experimental` module and a separate `tools` module.

Other notes:
 * This addresses #147 and #156.
 * This should go in after #159. I will resolve the conflicts.
 * This is groundwork for a bunch of code we'll be writing to prepare for the fall quarter (experimental features, rotators, new students, benchmarking).


pbailis:
@sahaana @kexinrong @edgan8 

@raininthesun: let's talk about this one, there are some issues with Contexts and I'm not sure whether you are planning to rewrite this or if I should put more time into fixing them. (Primarily, strange interactions between the tests and the `DatumEncoder`.

raininthesun:
@pbailis The context code has been stale. I will rebase/rewrite the whole thing once the algorithms are fixed. So you can leave the context code if that do not hold back the refactor.  Can you explain what is strange interactions between the tests and the DatumEncoder?

pbailis:
@kexinrong @edgan8: this is good to go.

pbailis:
@raininthesun: the errors have to with how we encode attributes. In the new code, I've written a helper class that converts generic `attribute`s into contextual attributes, but there's an off-by-one error in one of the tests. Let's see what happens after the rewrite.

kexinrong:
New project structure looks good to me. Quick question: why is `DatumWithNorm` under `macrobase.analysis.result` and `DatumWithScore` under `macrobase.analysis.summary.result`?

pbailis:
Thanks, @kexinrong: I think we can safely remove both of these classes. Working on it now.

pbailis:
@kexinrong: moved `DatumWithNorm` as a protected class of `OutlierClassifier` and removed `DatumWithScore` entirely.

pbailis:
Also pushed a fix that removes `targetLowMetrics` entirely and instead uses a FeatureTransform 😮 🎊 

Now all we have are `metrics` and `attributes`.

PR #161, opened 2016-08-16
Use attribute-based relative risk ratio.

pbailis:


pbailis:
@sahaana @kexinrong

core/src/main/java/macrobase/analysis/summary/itemset/RiskRatio.java
```diff
@@ -0,0 +1,62 @@
...
+
+        // no exposure occurred
+        if (totalExposedCount == 0) {
+            log.error("Computing risk ratio with no exposure.");
+            return 0;
+        }
+
+        // we only exposed this ratio, everything matched!
+        if (totalMinusExposedCount == 0) {
+            return 0;
```
kexinrong: It's not clear to me why when everything is exposed, the risk ratio should be 0.

kexinrong: One common solution to zero events seems to be adding 0.5 to all cell values (e.g. http://www.ncbi.nlm.nih.gov/pubmed/8085071). I think it might be worth spending a little time thinking about the edge cases.

pbailis: Let's discuss this tomorrow. Our thinking was that, if everything is exposed, it's not clear the utility the RR will have; specifically, in the case of MB queries, it likely means you misconfigured your query. Compared to adding 0.5 ([also](http://handbook.cochrane.org/chapter_16/16_9_2_studies_with_zero_cell_counts.htm)), we thought returning zero given the expected use cases would be more intuitive. But maybe we should just `log.warn` and do the 0.5 trick.

kexinrong:
LGTM!

PR #163, opened 2016-08-16
Display exceptions in GUI

pbailis:
Adds functionality in the GUI to display exceptions that occur while processing requests in the server.
example-err

Basic implementation idea is to wrap all responses with an extra class with a nullable `errorMessage` field.

### DIFF
--- a/bin/batch.sh
+++ b/bin/batch.sh
@@ -2,5 +2,4 @@
 conf_file=${1:-"conf/batch.yaml"}
 
 set -x
-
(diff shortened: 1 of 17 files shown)


pbailis:
@srmadden!

pbailis:
@sahaana  @kexinrong: good to go!

pbailis:
@kexinrong @sahaana tomorrow, perhaps?

kexinrong:
lgtm

PR #166, opened 2016-10-20
Allow configuring any pipeline to work with web GUI

mamikonyana:
This let's the user to specify a pipeline using
`macrobase.pipeline.class` same as during command line usage.

### DIFF
--- a/frontend/src/main/java/macrobase/runtime/resources/AnalyzeResource.java
+++ b/frontend/src/main/java/macrobase/runtime/resources/AnalyzeResource.java
@@ -2,15 +2,18 @@ package macrobase.runtime.resources;
 
 import macrobase.MacroBase;
 import macrobase.analysis.pipeline.BasicBatchedPipeline;
-import macrobase.conf.MacroBaseConf;
+import macrobase.analysis.pipeline.Pipeline;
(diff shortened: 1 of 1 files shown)


mamikonyana:

frontend/src/main/java/macrobase/runtime/resources/AnalyzeResource.java
```diff
@@ -59,7 +62,17 @@ public AnalysisResponse getAnalysis(AnalysisRequest request) {
                 conf.set(MacroBaseConf.DATA_LOADER_TYPE, MacroBaseConf.DataIngesterType.CSV_LOADER);
             }
 
-            List<AnalysisResult> results = new BasicBatchedPipeline().initialize(conf).run();
+
+            Class c = Class.forName(conf.getString(MacroBaseConf.PIPELINE_NAME, MacroBaseDefaults.DEFAULT_PIPELINE));
```
pbailis: Can we create a private final static member `defaultPipeline` inside `AnalyzeResource` and use this instead of adding it to `MacroBaseDefaults`?

mamikonyana:
@pbailis another look?

PR #167, opened 2016-10-20
Add requirements.txt

rajpurkar:
Add requirements file to install python dependency to have the tutorial running.


pbailis:
Thanks for this. We have a `requirements.txt` in `tools/`; is there any way to use that one instead?

rajpurkar:
Oops I didn't catch that! That definitely works, thanks!

Issue #170, opened 2016-11-01
Address already in use.


kemathur:
I was recently introduced to MacroBase and wanted to give it a try. I have been following the [tutorial](https://github.com/stanford-futuredata/macrobase/wiki/Tutorial) and I had followed the steps but when I run bin/server.sh I get an exception:
```

INFO  [2016-10-29 07:09:45,783] io.dropwizard.server.DefaultServerFactory: Registering admin handler with root path prefix: /
INFO  [2016-10-29 07:09:45,837] org.eclipse.jetty.setuid.SetUIDListener: Opened application@435871cb{HTTP/1.1}{0.0.0.0:8080}
WARN  [2016-10-29 07:09:45,841] org.eclipse.jetty.util.component.AbstractLifeCycle: FAILED org.eclipse.jetty.server.Server@45cff11c: java.lang.RuntimeException: java.net.BindException: Address already in use
! java.net.BindException: Address already in use
! at sun.nio.ch.Net.bind0(Native Method) ~[na:1.8.0_111]
! at sun.nio.ch.Net.bind(Net.java:433) ~[na:1.8.0_111]
! at sun.nio.ch.Net.bind(Net.java:425) ~[na:1.8.0_111]
! at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223) ~[na:1.8.0_111]
! at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74) ~[na:1.8.0_111]
! at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:321) ~[macrobase-jar-with-dependencies.jar:na]
! at org.eclipse.jetty.setuid.SetUIDListener.lifeCycleStarting(SetUIDListener.java:200) ~[macrobase-jar-with-dependencies.jar:na]
! ... 8 common frames omitted
! Causing: java.lang.RuntimeException: java.net.BindException: Address already in use
! at org.eclipse.jetty.setuid.SetUIDListener.lifeCycleStarting(SetUIDListener.java:213) ~[macrobase-jar-with-dependencies.jar:na]
! at org.eclipse.jetty.util.component.AbstractLifeCycle.setStarting(AbstractLifeCycle.java:188) ~[macrobase-jar-with-dependencies.jar:na]
[...]

PR #172, opened 2016-11-02
Shade assembly jar to include Dropwizard resources

pbailis:
Per https://groups.google.com/forum/#!topic/dropwizard-user/PE-k9wm7fXk

Needed to enable changing ports as in
```
server:
    applicationConnectors:
        - type: http
          port: 6666
    adminConnectors:
        - type: http
          port: 6667
```

IIRC, this previously worked. Not sure what changed...


pbailis:
Needed for #170

PR #174, opened 2016-11-17
prefer interface over abstract class

sweatybridge:


pbailis:
This is a reasonable suggestion but going to hold off for now since this is mostly stylistic.

PR #175, opened 2016-11-30
Winsorized Empirical Covariance transform

edgan8:
alternative to MCD: not as robust but 20x faster

It seems useful to have a simpler multi-variate outlier detector for quick runs on large datasets. MCD on 2Million points takes 35 seconds which is annoying, this new operator (RCOV) does not require multiple iterations like MCD and completes in 2.8 seconds while still proving results that are useful for downstream itemset mining.

Results on a 2d gaussian with offset uniform noise:
```
RCOV:

Result: [outliers: 21000.000000
inliers: 2079000.000000
load time 8959ms
execution time: 2819ms
summarization time: 164ms

-----

support: 0.969571
records: 20361.000000
ratio: 637.276995

Columns:
    class: B

MCD:
Result: [outliers: 21000.000000
inliers: 2079000.000000
load time 8965ms
execution time: 36709ms
summarization time: 445ms

-----

support: 0.986048
records: 20707.000000
ratio: 1413.447099

Columns:
    class: B

-----
```

A MAD on one of the metrics only has a ratio of 15, so RCOV is still picking up the important multivariate structure. The above results can be reproduced by running

```
python3 tools/gen_data/gen_multivariate_gaussian.py
./bin/batch.sh tools/demos/multivariate_gaussian/mcd.yaml
./bin/batch.sh tools/demos/multivariate_gaussian/rcov.yaml
./bin/batch.sh tools/demos/multivariate_gaussian/mad.yaml
```

So we see that RCOV is much faster but has slightly less "clean" results compared to MCD.

In terms of implementation, a lot of code is similar to the MCD code, but I think factoring out a specific module to just fit a Gaussian makes things easier to test and re-use. MCD can probably call the new Gaussian module in the future. The Gaussian fitting is made more robust to outliers by winsorizing input data independently along each metric such that the most extreme values are clipped to upper and lower percentile (by default 5%).


edgan8:
cc @pbailis @kexinrong @sahaana

core/src/main/java/macrobase/analysis/stats/RobustEmpiricalCovariance.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class RobustEmpiricalCovariance extends BatchTrainScore {
+    public Winsorizer trimmer;
+    public Gaussian gModel;
+
+    public static final String CONF_TRIM_PCT = "macrobase.analysis.rcov.trimPercent";
+    public static final double CONF_TRIM_PCT_DEFAULT = 5.0;
```
kexinrong: The default can go to MacroBaseDefaults?

kexinrong:
LGTM overall. And I agree that we should probably refactor MCD to use Gaussian after this gets merged.

PR #176, opened 2016-12-03
Batch window average

kexinrong:
The commit includes the following changes:
- Rename WINDOW_SIZE in MacroBaseConf to TIME_WINDOW, to make it clear that 
- Change Datum's time type from int to long, since int might be too small to express large time intervals in miliseconds
- Implemented batch window average

### DIFF
--- a/contrib/src/main/java/macrobase/analysis/transform/aggregate/BatchWindowAvg.java
+++ b/contrib/src/main/java/macrobase/analysis/transform/aggregate/BatchWindowAvg.java
@@ -1,0 +1,37 @@
+package macrobase.analysis.transform.aggregate;
+
+import macrobase.conf.ConfigurationException;
+import macrobase.conf.MacroBaseConf;
+import macrobase.conf.MacroBaseDefaults;
+import macrobase.datamodel.Datum;
+import org.apache.commons.math3.linear.ArrayRealVector;
+import org.apache.commons.math3.linear.RealVector;
+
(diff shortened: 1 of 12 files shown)


contrib/src/main/java/macrobase/analysis/transform/aggregate/BatchWindowAvg.java
```diff
@@ -0,0 +1,39 @@
...
+    public BatchWindowAvg(MacroBaseConf conf) throws ConfigurationException {
+        this.timeColumn = conf.getInt(MacroBaseConf.TIME_COLUMN, MacroBaseDefaults.TIME_COLUMN);
+    }
+
+    public Datum aggregate(List<Datum> data) {
+        if (data.isEmpty())
+            return new Datum(new ArrayList<>(), new ArrayRealVector(dim));
+
+        dim = data.get(0).metrics().getDimension();
+        RealVector results = new ArrayRealVector(dim);
```
edgan8: Doesn't seem like we benefit much from using a RealVector here over a double[], double[] is going to be faster (can just create an ArrayRealVector from it in the end)

kexinrong: Yes -- ASAP is going to use BatchWindowAvg as the transform. Right now we only have BatchWindowMax.
And sounds good. I will switch to using double[].

PR #178, opened 2016-12-05
Tkde foundations

edgan8:
This commit includes code for kernels, kdtree, and a naive kde.
It duplicates some functionality in the existing kde code but this
implementation is much faster and I believe better factored out into
components. Some conventions that I think will be useful going forward:

1) Separate functionality into self-contained mathematical modules which
are independent of macrobase and can be tested independently of
macrobase, and then combine them at the end into a mb operator. e.g.
bandwidth selectors, kernels, and even kde. Configuration can be done
via builder pattern for better composability rather than having each
component take a config object.

2) Tests can load numeric data from a csv directly via new helper
utilities rather than going through the heavy mb csv ingester.

3) Using raw double[] rather than datum or vector yields substantial
improvements in runtime.


pbailis:
Closing due to inactivity; see also https://github.com/stanford-futuredata/tKDC

PR #188, opened 2017-01-26
Remove redirect to # on 'Explore' click

pbailis:
One-liner:

Old setting: the link target for "explore" redirected to "#"
Old behavior: if you click "explore", then tab back to the console, you lose your place on the page

New setting: the link target for "explore" is not defined
New behavior: if you click "explore", then tab back to the console, you keep your place

Technically, omitting the `href` tag is bad HTML, but the app doesn't even work without javascript.

### DIFF
--- a/frontend/src/main/resources/frontend/console.html
+++ b/frontend/src/main/resources/frontend/console.html
@@ -234,7 +234,7 @@
                                 </div>
                             </div>
                             <div style="padding: 15px; padding-top:0px;">
-                                <a href="#" ng-click="exploreItems(i.items)">Explore</a>
+                                <a ng-click="exploreItems(i.items)">Explore</a>
                             </div>
                         </div>
                     </div>


edgan8:
Does this need to be a link then? Can it just be a styled div?

pbailis:
@edgan8: so, technically this acts just like a link. We need to capture the context in which the user pressed "explore" to pass it to the actual explorer module in angular, so we use JS then do a window redirect. This is kind of a hack, and I'm sure there's a more idiomatic way to do so (at least I hope so for the sake of JS; maybe you pass the params as part of the redirect?) but it's what we have now for passing the attribute-value pairs to the explorer view. So, given that it's basically a link, just with a weird JS-magic target, I'm in favor of keeping it simple as a link for now. Happy to throw some UX magic on it, possibly this summer? Or maybe we rewrite it in angular2?

pbailis:
@sahaana thanks!

PR #189, opened 2017-02-01
Allow overweight items in AChao sampler

pbailis:
Two fixes to the core reservoir sampler:

1. p(insertion) should be (res_size * weight) / (total weight in res). Before, we only had (weight) / (total weigh in res), which is incorrect.

2. We haven't been handling "overweight" items, where (res_size * weight) / (total weight) > 1. The right thing to do here ([source](https://arxiv.org/pdf/1012.0256.pdf)) is to keep all overweight items in the reservoir. The way I've implemented it is to have the overweight items in their own priority queue. Then, when a user wants to query the reservoir, if there are overweight items, they're merged into the reservoir. When we decay the reservoir and/or insert new items (thus updating the weights), the overweight items are checked (again, in order of weight) to check if they're no longer overweight. If they are no longer overweight, we insert them like a normal item into the main reservoir.

This sounds more complicated than it actually is. The code is pretty simple.

This situation should be _exceedingly_ rare in practice and really only kicks in when 1.) users insert enormously-weighted items into the reservoir and/or 2.) users do a ton of exponential decay. I don't think we've hit either in a real scenario. But, I wanted to do it right, just in case.

### DIFF
--- a/core/src/main/java/macrobase/analysis/sample/AChao.java
+++ b/core/src/main/java/macrobase/analysis/sample/AChao.java
@@ -1,18 +1,45 @@
 package macrobase.analysis.sample;
 
+import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
+import java.util.PriorityQueue;
 import java.util.Random;
+import java.util.stream.Collectors;
 
 /**
  * See http://arxiv.org/pdf/1012.0256.pdf
  */
 class AChao<T> {
+    private class OverweightItem<T> implements Comparable<OverweightItem> {
+        public Double weight;
+        public T item;
+
+        public OverweightItem(T item,
+                              double weight) {
+            this.item = item;
+            this.weight = weight;
+        }
+
+        @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/macrobase/analysis/sample/AChao.java
```diff
@@ -24,17 +51,56 @@ public AChao(int capacity, Random random) {
...
+    }
+
     public final List<T> getReservoir() {
+        updateOverweightItems();
+
+        if(!overweightItems.isEmpty()) {
+            // overweight items always make it in the sample
+            List<T> ret = overweightItems.stream().map(i -> i.item).collect(Collectors.toList());
+
+            assert (ret.size() <= reservoirCapacity);
```
edgan8: Why is this guaranteed? Seems like the list could be very big, might as well take the first n elements.

pbailis: The answer has to do with updateOverweightItems...

There are at most reservoirCapacity overweight items at any time.

edgan8: wait can you point out which line that logic is in? I guess I can't find it.

pbailis: Ah, sorry, yeah, this should be better documented. (not an excuse, but it is also really hard to grok in the paper).

`updateOverweightItems` has the logic that will do the pruning.

I missed the above stat test etc., but yeah, we should do that :)

PR #191, opened 2017-02-23
simple fix for boxed NPE

pbailis:
One-liner NPE fix; goes in after #189


core/src/main/java/macrobase/analysis/summary/itemset/FPGrowthEmerging.java
```diff
@@ -150,7 +150,7 @@ public FPGrowthEmerging(boolean combinationsEnabled) {
             if (i.getItems().size() == 1) {
                 Number inlierCount = inlierCounts.get(i.getItems().iterator().next());
 
-                double ratio = RiskRatio.compute(inlierCount.doubleValue(),
+                double ratio = RiskRatio.compute(inlierCount,
```
edgan8: is handling nulls being delegated to RiskRatio.compute here?

pbailis: yeah, we call RR from a number of places, so RR does the checks

PR #195, opened 2017-04-12
Incremental Summarizer Implementation

kexinrong:


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * Given a batch of rows with an outlier class column, explain the outliers using
+ * string attribute columns. Each batch is considered as an independent unit.
+ */
+public class IncrementalSummarizer implements IncrementalOperator<Explanation> {
+    // pane size known in advance
+    public int numPanes;
+
+    // Parameters
+    protected String outlierColumn = "_OUTLIER";
```
pbailis: is this a default? if so, should say so.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+    protected double minOutlierSupport = 0.1;
+    protected double minRiskRatio = 3;
+    protected boolean useAttributeCombinations = true;
+    protected List<String> attributes = new ArrayList<>();
+    protected DoublePredicate predicate = d -> d != 0.0;
```
pbailis: ditto above.

could use a comment on what predicate does.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+
+    // Encoder
+    protected AttributeEncoder encoder = new AttributeEncoder();
```
pbailis: I thought this was deprecated?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+    private HashMap<Set<Integer>, Double> outlierItemsetPaneCount = new HashMap<>();
+    private HashMap<Set<Integer>, Double> inlierItemsetWindowCount = new HashMap<>();
+    private HashMap<Set<Integer>, Double> outlierItemsetWindowCount = new HashMap<>();
+    private Deque<Integer> inlierPaneCounts;
+    private Deque<Integer> outlierPaneCounts;
+    private List<Integer> inlierCountCumSum;
+    private List<Integer> outlierCountCumSum;
+    private HashMap<Set<Integer>, Integer> trackingMap = new HashMap<>();
+
+    public IncrementalSummarizer(int numPanes) { setWindowSize(numPanes); }
```
pbailis: comment describing implementation strategy would be very helpful.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @Override
+    public void setWindowSize(int numPanes) {
```
pbailis: why not at construction time? what is the intended scenario where window size changes?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+            outlierPaneCounts = new ArrayDeque<>(numPanes);
+            inlierItemsetPaneCounts = new ArrayDeque<>(numPanes);
+            outlierItemsetPaneCounts = new ArrayDeque<>(numPanes);
+        }
+    }
+
+    @Override
+    public int getWindowSize() { return numPanes; }
+
+    public IncrementalSummarizer setMinSupport(double minSupport) {
```
pbailis: ditto here, why not set at construction time? what is the intended scenario where support will change? what are the semantics if in fact support changes mid-way?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+        this.minOutlierSupport = minSupport;
+        return this;
+    }
+
+    /**
+     * Adjust this to tune the severity (e.g. strength of correlation) of the results returned.
+     * @param minRiskRatio lowest risk ratio to consider for meaningful explanations.
+     * @return this
+     */
+    public IncrementalSummarizer setMinRiskRatio(double minRiskRatio) {
```
pbailis: can this just be a parameter we pass at mine() time?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+        return this;
+    }
+
+    /**
+     * Whether or not to use combinations of attributes in explanation, or only
+     * use simple single attribute explanations
+     * @param useAttributeCombinations flag
+     * @return this
+     */
+    public IncrementalSummarizer setUseAttributeCombinations(boolean useAttributeCombinations) {
```
pbailis: is this functionality we want to expose? i had it in macrobasev0 as a hack... happy to keep it, but curious if this is a requested use case.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+        } else {
+            encoder.setColumnNames(attributes);
+            inlierItemsets = encoder.encodeAttributes(inlierDF.getStringColsByName(attributes));
+            outlierItemsets = encoder.encodeAttributes(outlierDF.getStringColsByName(attributes));
+        }
+        inlierPaneCounts.add(inlierItemsets.size());
+        outlierPaneCounts.add(outlierItemsets.size());
+    }
+
+    private void calcCumSum() {
```
pbailis: add comment for this function... what is it doing? where is it called? why is it protected?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+        outlierCountCumSum = new ArrayList<>(inlierPaneCounts.size() + 1);
+        inlierCountCumSum.add(0);
+        outlierCountCumSum.add(0);
+        for (int i = 0; i < inlierPaneCounts.size(); i ++) {
+            inlierCountCumSum.add(inlierCountCumSum.get(i) + (int)inlierCounts[i]);
+            outlierCountCumSum.add(outlierCountCumSum.get(i) + (int)outlierCounts[i]);
+        }
+    }
+
+    private void removePaneCounts() {
```
pbailis: ditto, function;

what about "expireLastPane"?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+        // Decrement pane numbers that itemsets first got promoted since we removed the first pane
+        for (Set<Integer> itemset : trackingMap.keySet()) {
+            int paneNo = trackingMap.get(itemset);
+            if (paneNo > 0) {
+                trackingMap.put(itemset, paneNo - 1);
+            }
+        }
+    }
+
+    private void addPaneCounts() {
```
pbailis: ditto, comment here...

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+        inlierItemsetWindowCount.keySet().removeAll(unSupported);
+    }
+
+    private void trackItemset(Set<Integer> itemset, double count) {
+        trackingMap.put(itemset, inlierPaneCounts.size() - 1);
+        outlierItemsetWindowCount.put(itemset, count);
+        outlierItemsetPaneCount.put(itemset, count);
+    }
+
+    private void addNewFrequent() {
```
pbailis: explain when/why this is used?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void process(DataFrame df) {
+        // 1. Retire old pane counts if necessary
+        if (inlierPaneCounts.size() == numPanes) { removePaneCounts(); }
+        // 2. Add support counts for the new pane
+        encodeAttributes(df);
```
pbailis: this is destructive; are the semantics for `process` always that dfs consumed should not be reused?

kexinrong: This function doesn't modify the input dataframe -- it just produces the encoded inlier and outlier itemsets. But I agree the function signature is kind of confusing.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
kexinrong: I just inherited this from the exisiting code base. I don't think we have a use case for this so happy to remove it.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
kexinrong: One scenario is that the user wants to change parameters of the summarizer in a long running job, but don't want to start a new class and discard all existing results. In this case, we start using the new parameter for future data. There will be some mixed results in between, but eventually, results under the old parameter setting will retire and we are able to fully convert to the new setting.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,311 @@
...
+import edu.stanford.futuredata.macrobase.operator.IncrementalOperator;
+
+import java.util.*;
+import java.util.function.DoublePredicate;
+
+/**
+ * Given batches of rows with an outlier class column, explain the outliers using
+ * string attribute columns. Results from each batch accumulates, until batches are retired.
+ */
+public class IncrementalSummarizer implements IncrementalOperator<Explanation> {
```
pbailis: Minor: Do we want to keep "Summarizer" or just "Explainer"?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,311 @@
...
+    //
```
pbailis: missing comment? 

getters and setters instead of public

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,311 @@
...
+    protected double minOutlierSupport = 0.1;
+    protected List<String> attributes = new ArrayList<>();
+    // Default predicate for filtering outlying rows
+    protected DoublePredicate predicate = d -> d != 0.0;
+
+    // Encoder and encoded attribute sets
+    protected AttributeEncoder encoder = new AttributeEncoder();
+    protected List<Set<Integer>> inlierItemsets, outlierItemsets;
+
+    // Internal booking keep
```
pbailis: bookkeeping

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,311 @@
...
+    private Deque<HashMap<Set<Integer>, Double>> outlierItemsetPaneCounts;
+    private HashMap<Set<Integer>, Double> inlierItemsetPaneCount = new HashMap<>();
+    private HashMap<Set<Integer>, Double> outlierItemsetPaneCount = new HashMap<>();
+    private HashMap<Set<Integer>, Double> inlierItemsetWindowCount = new HashMap<>();
+    private HashMap<Set<Integer>, Double> outlierItemsetWindowCount = new HashMap<>();
+    private Deque<Integer> inlierPaneCounts;
+    private Deque<Integer> outlierPaneCounts;
+    private List<Integer> inlierCountCumSum;
+    private List<Integer> outlierCountCumSum;
+    private HashMap<Set<Integer>, Integer> trackingMap = new HashMap<>();
```
pbailis: quick comment?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
pbailis: Hmm, fair. Kind of awkward, but I suppose if the alternative is to replay the stream, it makes sense.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,311 @@
...
+     * @param outlierColumn new outlier indicator column.
+     * @return this
+     */
+    public IncrementalSummarizer setOutlierColumn(String outlierColumn) {
+        this.outlierColumn = outlierColumn;
+        return this;
+    }
+
+    /* Encode inlier and outlier attributes into itemsets */
+    private void encodeAttributes(DataFrame df) {
```
pbailis: ideally we'd get rid of encoding entirely, but fine for now.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/IncrementalSummarizer.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    @Override
+    public void process(DataFrame df) {
+        // 1. Retire old pane counts if necessary
+        if (inlierPaneCounts.size() == numPanes) { expireLastPane(); }
+        // 2. Add support counts for the new pane
+        encodeAttributes(df);
+        addNewPane();
+        if (outlierItemsets.size() * minOutlierSupport >= 1) {
```
edgan8: Clean up the logic here: I think pruning should always happen, for addNewFrequent we may need to be more careful.

kexinrong: Moved this to addNewFrequent(). This is saying that if there are too few outliers in the result, we don't promote new frequent itemset (since all outlier itemset would be promoted this way).

edgan8:
OK let's commit this, we can run some more runtime / memory benchmarks on monday before pushing to maven

PR #197, opened 2017-04-25
Metric Bucket Transformer

edgan8:
Transform metrics to attributes to explain metrics using correlated other metrics. Buckets floating point value by percentiles, converting them to string attributes that are tagged with either the bucket range or the bucket index.

See MetricAsExplanationTest for complete usage. Using smaller buckets allows more precise explanations with higher risk ratio.

### DIFF
--- a/lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/transform/MetricBucketTransformer.java
+++ b/lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/transform/MetricBucketTransformer.java
@@ -6,16 +6,31 @@ import org.apache.commons.math3.stat.descriptive.rank.Percentile;
 
 import java.util.*;
 
+/**
+ * Transform real valued columns into categorical string columns.
+ * Transformed columns are added to a copy of the dataframe and appended with
+ * a suffix to distinguish them from the original column.
+ * Useful for using correlated metrics as explanatory values.
+ * By default transforms columns by bucketing them into low-med-high values
+ * based on percentile.
+ */
 public class MetricBucketTransformer implements Transformer {
+    // Transformed columns are added to the dataframe under a suffix.
     private String columnSuffix = "_a";
+    // Boundaries of the buckets used for classification. n boundaries -> n+1 buckets
     private double[] boundaryPercentiles = {10.0, 90.0};
-    private boolean useSimpleNames = false;
+    // The strings used to encode which bucket a value falls in can be either a simple index
+    // or an encoding of the range of the bucket.
+    private boolean simpleBucketValues = false;
 
     private List<String> metricColumns;
     private List<String> transformedColumnNames;
 
     private DataFrame transformedDF;
 
+    /**
+     * @param columns set of columns to transform
+     */
     public MetricBucketTransformer(List<String> columns) {
         this.metricColumns = columns;
         int d = columns.size();
@@ -47,7 +62,7 @@ public class MetricBucketTransformer implements Transformer {
             }
 
             String[] bucketNames = new String[k+1];
-            if (useSimpleNames) {
+            if (simpleBucketValues) {
                 for (int i = 0; i < k+1; i++) {
                     bucketNames[i] = String.format("%s:%d", colName, i);
                 }
@@ -95,12 +110,12 @@ public class MetricBucketTransformer implements Transformer {
         this.boundaryPercentiles = boundaryPercentiles;
     }
 
-    public boolean isUseSimpleNames() {
-        return useSimpleNames;
+    public boolean isSimpleBucketValues() {
+        return simpleBucketValues;
     }
 
-    public void setUseSimpleNames(boolean useSimpleNames) {
-        this.useSimpleNames = useSimpleNames;
+    public void setSimpleBucketValues(boolean simpleBucketValues) {
+        this.simpleBucketValues = simpleBucketValues;
     }
 
     public List<String> getTransformedColumnNames() {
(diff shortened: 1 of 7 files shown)


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/transform/MetricBucketTransformer.java
```diff
@@ -0,0 +1,109 @@
+package edu.stanford.futuredata.macrobase.analysis.transform;
+
+import edu.stanford.futuredata.macrobase.datamodel.DataFrame;
+import edu.stanford.futuredata.macrobase.operator.Transformer;
+import org.apache.commons.math3.stat.descriptive.rank.Percentile;
+
+import java.util.*;
+
+public class MetricBucketTransformer implements Transformer {
```
pbailis: Short comment to explain class purpose and behavior would be very useful.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/transform/MetricBucketTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+    private String columnSuffix = "_a";
+    private double[] boundaryPercentiles = {10.0, 90.0};
+    private boolean useSimpleNames = false;
```
pbailis: Comment

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/transform/MetricBucketTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public double[] getBoundaryPercentiles() {
+        return boundaryPercentiles;
+    }
+
+    public void setBoundaryPercentiles(double[] boundaryPercentiles) {
+        this.boundaryPercentiles = boundaryPercentiles;
+    }
+
+    public boolean isUseSimpleNames() {
```
pbailis: style nit: kind of awk for `isUse` vs `use`

lib/src/main/java/edu/stanford/futuredata/macrobase/datamodel/DataFrame.java
```diff
@@ -9,7 +9,7 @@
 
 /**
  * Column-based dataframe object.
- * loadRows and addColumn methods mutate the dataframe and are the primary
+ * addColumn methods mutate the dataframe and are the primary
  * ways of initializing the data in the dataframe.
```
pbailis: ways -> means?

lib/src/main/java/edu/stanford/futuredata/macrobase/datamodel/Row.java
```diff
@@ -55,11 +55,16 @@ public int hashCode() {
...
-                if (s.length() > 7) {
-                    double v = ((Double) o).doubleValue();
-                    s = String.format("%.6g", v);
+            String s;
+            if (o == null) {
+                s = "NULL";
+            } else {
+                s = o.toString();
+                if (o instanceof Double) {
+                    if (s.length() > 7) {
```
pbailis: Kind of arbitrary, but okay... Any reason why 7?

lib/src/main/java/edu/stanford/futuredata/macrobase/operator/Operator.java
```diff
@@ -1,5 +1,13 @@
 package edu.stanford.futuredata.macrobase.operator;
 
+/**
+ * Generic interface for operations. Should not modify input.
```
pbailis: Previously the semantics were to allow operators to destroy inputs; otherwise we have to copy everything on ever operation. What's the rationale for changing this?

edgan8: Now that we have a column store, creating a shallow "copy" of a dataframe with extra or missing columns is cheap and makes the code easier to reason about. Adding or removing rows in a column store is going to require a new copy anyway so no loss there either.

Is there a specific scenario you have in mind where this hurts us?

pbailis: Understood; wasn't thinking about column storage.

However, I _was_ thinking of in-place transforms to metrics; won't that be expensive? (Probably overthinking here.)

edgan8: Most transforms I can think of are going to modify a significant fraction of the column values, in which case creating a new column isn't much overhead on top of computing these values.

If we come across an operation that really benefits from in-place mutation I figure we can create a new "Mutator" interface to make the distinction very explicit.

lib/src/main/java/edu/stanford/futuredata/macrobase/datamodel/Row.java
edgan8: Just adjusted it until the floating point numbers were a reasonable length, 1.3e-10 is already 7 characters.

pbailis: works for me.

lib/src/main/java/edu/stanford/futuredata/macrobase/operator/Operator.java
pbailis: ack! i like the "Mutator" idea.

pbailis:
Good to go @edgan8

PR #198, opened 2017-05-25
Add MultiMAD classifier

jialinding:
I added a MultiMADClassifier class and also a unit test. I modeled it after the PercentileClassifier, so switching between them shouldn't be difficult. I have some graphs to back up my implementation decisions, which we can discuss in person.

### DIFF
--- a/lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
+++ b/lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
@@ -1,5 +1,6 @@
 package edu.stanford.futuredata.macrobase.analysis.classify;
 
+import edu.stanford.futuredata.macrobase.analysis.sample.*;
 import edu.stanford.futuredata.macrobase.datamodel.DataFrame;
 import edu.stanford.futuredata.macrobase.operator.Transformer;
 
@@ -26,16 +27,22 @@ public class MultiMADClassifier implements Transformer {
     private boolean includeHigh = true;
     private boolean includeLow = true;
     private double samplingRate = 1;
+    private Sampler sampler = null;
+    private boolean useReservoirSampling = false;
     private List<String> columnNames;
     private String outputColumnSuffix = "_OUTLIER";
     private boolean useParallel = false;
-    private boolean useReservoirSampling = false;
+    
     private final double trimmedMeanFallback = 0.05;
     // https://en.wikipedia.org/wiki/Median_absolute_deviation#Relation_to_standard_deviation
     private final double MAD_TO_ZSCORE_COEFFICIENT = 1.4826;
 
     // Calculated values
     private DataFrame output;
+
+    public MultiMADClassifier(List<String> columnNames) {
+        this.columnNames = columnNames;
+    }
 
     public MultiMADClassifier(String... columnNames) {
         this.columnNames = new ArrayList<String>(Arrays.asList(columnNames));
@@ -45,25 +52,14 @@ public class MultiMADClassifier implements Transformer {
     public void process(DataFrame input) {
         output = input.copy();
 
-        int sampleSize = (int)(input.getNumRows() * samplingRate);
-        int[] sampleIndices = new int[sampleSize];
         if (samplingRate != 1) {
-            sampleIndices = getRandomSample(input.getNumRows());
+            sampler.computeSampleIndices(input.getNumRows(), samplingRate);
         }
 
         for (String columnName : columnNames) {
             double[] metrics = input.getDoubleColumnByName(columnName);
 
-            double[] trainingMetrics = new double[sampleSize];
-            if (samplingRate == 1) {
-                System.arraycopy(metrics, 0, trainingMetrics, 0, metrics.length);
-            } else {
-                for (int j = 0; j < sampleIndices.length; j++) {
-                    trainingMetrics[j] = metrics[sampleIndices[j]];
-                }
-            }
-
-            Stats stats = train(trainingMetrics);
+            Stats stats = train(metrics);
             double lowCutoff = stats.median - (zscore * stats.mad);
             double highCutoff = stats.median + (zscore * stats.mad);
 
@@ -87,60 +83,37 @@ public class MultiMADClassifier implements Transformer {
         }
     }
 
-    private int[] getRandomSample(int numRows) {
-        int sampleSize = (int)(numRows * samplingRate);
-        int[] sampleIndices = new int[sampleSize];
-
-        if (useReservoirSampling) {
-            // Reservoir Sampling:
-            for (int i = 0; i < sampleSize; i++) {
-                sampleIndices[i] = i;
-            }
-            Random rand = new Random();
-            for (int i = sampleSize; i < numRows; i++) {
-                int j = rand.nextInt(i+1);
-                if (j < sampleSize) {
-                    sampleIndices[j] = i;
-                }
-            }
-            boolean[] mask = new boolean[numRows];
-            for (int i = 0; i < sampleSize; i++) {
-                mask[sampleIndices[i]] = true;
-            }
-        } else {
-            // partial Fisher-Yates shuffle
-            int[] range = new int[numRows];
-            for (int i = 0; i < numRows; i++) {
-                range[i] = i;
-            }
-            Random rand = new Random();
-            for (int i = 0; i < sampleSize; i++) {
-                int j = rand.nextInt(numRows - i) + i;
-                int temp = range[j];
-                range[j] = range[i];
-                range[i] = temp;
-            }
-            System.arraycopy(range, 0, sampleIndices, 0, sampleSize);
-        }
-
-        return sampleIndices;
-    }
-
     private Stats train(double[] metrics) {
-        double median = new Percentile().evaluate(metrics, 50);
+        Percentile percentile = new Percentile();
 
-        for (int i = 0; i < metrics.length; i++) {
-            metrics[i] = Math.abs(metrics[i] - median);
+        double[] trainingMetrics;  // used to compute the median
+        if (samplingRate == 1) {
+            trainingMetrics = metrics;
+        } else {
+            trainingMetrics = sampler.getSample(metrics);
         }
 
-        double MAD = new Percentile().evaluate(metrics, 50);
+        double median = percentile.evaluate(trainingMetrics, 50);
 
+        double[] residuals;  // used to compute the MAD
+        if (samplingRate == 1) {
+            residuals = new double[trainingMetrics.length];
+        } else {
+            residuals = trainingMetrics;
+        }
+        for (int i = 0; i < trainingMetrics.length; i++) {
+            residuals[i] = Math.abs(trainingMetrics[i] - median);
+        }
+
+        double MAD = percentile.evaluate(residuals, 50);
+
+        // If MAD is 0, we use the average of residuals
         if (MAD == 0) {
-            Arrays.sort(metrics);
-            int lowerTrimmedMeanIndex = (int) (metrics.length * trimmedMeanFallback);
-            int upperTrimmedMeanIndex = (int) (metrics.length * (1 - trimmedMeanFallback));
+            Arrays.sort(residuals);
+            int lowerTrimmedMeanIndex = (int) (residuals.length * trimmedMeanFallback);
+            int upperTrimmedMeanIndex = (int) (residuals.length * (1 - trimmedMeanFallback));
             double sum = 0;
-            for (int i = lowerTrimmedMeanIndex; i < upperTrimmedMeanIndex; ++i) {
+            for (int i = residuals.length/2; i < upperTrimmedMeanIndex; ++i) {
                 sum += metrics[i];
             }
             MAD = sum / (upperTrimmedMeanIndex - lowerTrimmedMeanIndex);
@@ -221,6 +194,13 @@ public class MultiMADClassifier implements Transformer {
      */
     public MultiMADClassifier setSamplingRate(double samplingRate) {
         this.samplingRate = samplingRate;
+        if (samplingRate != 1 && this.sampler == null) {
+            if (useReservoirSampling) {
+                this.sampler = new ReservoirSampler();
+            } else {
+                this.sampler = new FisherYatesSampler();
+            }
+        }
         return this;
     }
 
@@ -261,11 +241,23 @@ public class MultiMADClassifier implements Transformer {
      */
     public MultiMADClassifier setUseReservoirSampling(boolean useReservoirSampling) {
         this.useReservoirSampling = useReservoirSampling;
+        if (sampler != null) {
+            if (useReservoirSampling && !sampler.getSamplingMethod().equals("reservoir")) {
+                sampler = new ReservoirSampler();
+            } else if (!useReservoirSampling && sampler.getSamplingMethod().equals("reservoir")) {
+                sampler = new FisherYatesSampler();
+            }
+        }
         return this;
     }
 
     public MultiMADClassifier setColumnNames(String... columnNames) {
         this.columnNames = new ArrayList<String>(Arrays.asList(columnNames));
+        return this;
+    }
+
+    public MultiMADClassifier setColumnNames(List<String> columnNames) {
+        this.columnNames = columnNames;
         return this;
     }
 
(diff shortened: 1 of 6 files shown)


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,292 @@
...
+     * @param useParallel Whether to use parallelization, can be beneficial for large datasets
+     * @return this
+     */
+    public MultiMADClassifier setUseParallel(boolean useParallel) {
+        this.useParallel = useParallel;
+        return this;
+    }
+
+    /**
+     * @param use ReservoirSampling Whether to sample using reservoir sampling. Otherwise,
```
kexinrong: Since you took out the magic number in the original code, I would probably add one or two more sentences here on when it's good to use reservoir sampling VS Fisher-Yates.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,292 @@
...
+
+import java.lang.Math;
+
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Random;
+import java.util.stream.IntStream;
+
+public class MultiMADClassifier implements Transformer {
```
kexinrong: Might be helpful to add a comment here on how this classifier works, e.g. it classifies each input column independently, and adds a column of results for each input column etc.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,292 @@
...
+    private List<String> columnNames;
+    private String outputColumnSuffix = "_OUTLIER";
+    private boolean useParallel = false;
+    private boolean useReservoirSampling = false;
+    private final double trimmedMeanFallback = 0.05;
+    // https://en.wikipedia.org/wiki/Median_absolute_deviation#Relation_to_standard_deviation
+    private final double MAD_TO_ZSCORE_COEFFICIENT = 1.4826;
+
+    // Calculated values
+    private double lowCutoff;
```
kexinrong: Since there is no getter for lowCutoff and highCutoff, maybe remove them from class variables?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,292 @@
...
+    private final class Stats {
+        private final double median;
+        private final double mad;
+
+        public Stats(double median, double mad) {
+            this.median = median;
+            this.mad = mad;
+        }
+
+        public double getMedian() {
```
kexinrong: Remove getMedian() and getMAD()? Since you are not using them

kexinrong:
LGTM overall. Minor comments inline.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+    private boolean useParallel = false;
+    private boolean useReservoirSampling = false;
+    private final double trimmedMeanFallback = 0.05;
+    // https://en.wikipedia.org/wiki/Median_absolute_deviation#Relation_to_standard_deviation
+    private final double MAD_TO_ZSCORE_COEFFICIENT = 1.4826;
+
+    // Calculated values
+    private DataFrame output;
+
+    public MultiMADClassifier(String... columnNames) {
```
edgan8: Let's make a constructor which takes in a list as well: it's easier to work with lists when programmatically calling this.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+        this.columnNames = new ArrayList<String>(Arrays.asList(columnNames));
+    }
+
+    @Override
+    public void process(DataFrame input) {
+        output = input.copy();
+
+        int sampleSize = (int)(input.getNumRows() * samplingRate);
+        int[] sampleIndices = new int[sampleSize];
```
edgan8: Is this necessary (allocating a new array) when sampleRate == 1.0? might as well avoid extra work. In general I think it would be nicer if you could factor out all of the "sampling" logic more cleanly rather than mixing it inline in multiple places, see what you can do.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+        if (samplingRate != 1) {
+            sampleIndices = getRandomSample(input.getNumRows());
+        }
+
+        for (String columnName : columnNames) {
+            double[] metrics = input.getDoubleColumnByName(columnName);
+
+            double[] trainingMetrics = new double[sampleSize];
+            if (samplingRate == 1) {
+                System.arraycopy(metrics, 0, trainingMetrics, 0, metrics.length);
```
edgan8: Let's avoid making copies as much as possible, and when copies are required let's do it as close to the "usage" of the copy as possible. It seems like this copy is used because you want to sort the array later:
a) if you really need to sort it let's copy it right before the sort
b) let's just use the percentile class rather than sorting

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+            }
+
+            Stats stats = train(trainingMetrics);
+            double lowCutoff = stats.median - (zscore * stats.mad);
+            double highCutoff = stats.median + (zscore * stats.mad);
+
+            if (useParallel) {
+                // Parallelized stream scoring:
+                double[] results = new double[metrics.length];
+                System.arraycopy(metrics, 0, results, 0, metrics.length);
```
edgan8: This arraycopy seems fine: the mutation is right here as well

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+                    boolean isOutlier = (metrics[r] > highCutoff && includeHigh) ||
+                        (metrics[r] < lowCutoff && includeLow);
+                    results[r] = isOutlier ? 1.0 : 0.0;
+                }
+                output.addDoubleColumn(columnName + outputColumnSuffix, results);
+            }
+        }
+    }
+
+    private int[] getRandomSample(int numRows) {
```
edgan8: Let's factor this into it's own class (or two classes! one for fisher yates one for reservoir with a shared interface):
a) it will be generally useful
b) then the unit test is more self contained

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+        int sampleSize = (int)(numRows * samplingRate);
+        int[] sampleIndices = new int[sampleSize];
+
+        if (useReservoirSampling) {
+            // Reservoir Sampling:
+            for (int i = 0; i < sampleSize; i++) {
+                sampleIndices[i] = i;
+            }
+            Random rand = new Random();
```
edgan8: We should factor out all the "Random" generators used, ideally share them within the same class. That way we can optionally set seeds in the constructor for deterministic behavior

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+        double median = new Percentile().evaluate(metrics, 50);
+
+        for (int i = 0; i < metrics.length; i++) {
+            metrics[i] = Math.abs(metrics[i] - median);
+        }
+
+        double MAD = new Percentile().evaluate(metrics, 50);
+
+        if (MAD == 0) {
+            Arrays.sort(metrics);
```
edgan8: Would prefer using percentile, did you notice any performance improvement using sort?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+            int lowerTrimmedMeanIndex = (int) (metrics.length * trimmedMeanFallback);
+            int upperTrimmedMeanIndex = (int) (metrics.length * (1 - trimmedMeanFallback));
+            double sum = 0;
+            for (int i = lowerTrimmedMeanIndex; i < upperTrimmedMeanIndex; ++i) {
+                sum += metrics[i];
+            }
+            MAD = sum / (upperTrimmedMeanIndex - lowerTrimmedMeanIndex);
```
edgan8: Honestly it's not clear to me that this algorithm is particularly meaningful when MAD = 0. This class is designed to give a reasonably interpretable result as fast as possible: in that case for the degenerate case when MAD = 0 the z-scores will be totally messed up anyway, in this "backup" code path how about just setting MAD = (p(10) + p(90)) / 2 or something

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,291 @@
...
+     * "a" + outputColumnSuffix.
+     * @return this
+     */
+    public MultiMADClassifier setOutputColumnSuffix(String outputColumnSuffix) {
+        this.outputColumnSuffix = outputColumnSuffix;
+        return this;
+    }
+
+    private final class Stats {
+        private final double median;
```
edgan8: Just make this public lol

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
kexinrong: I believe this is the backup code in the original MAD class. Maybe we should change that at some point too.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
jialinding: The original implementation in legacy for the MAD = 0 case was to sum all residuals between the 5th and 95th percentiles, which requires actually sorting the residuals.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
jialinding: Given that the point of MAD is to be robust to outliers, something like MAD = (p(10) + p(90)) / 2 wouldn't be good, because if p(90) is an outlier, then MAD is much higher than it should be. The current implementation finds the average of all residuals between the 5th and 95th percentiles, which is more robust, but I guess still not ideal.

pbailis:
@jialinding this is good to go, right?

pbailis:
@jialinding does this need anything else?

jialinding:
@edgan8 Is this good to go? I wonder if it's worth it to add a pipeline that uses the MultiMADClassifier to identify the most interesting metrics.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/MultiMADClassifier.java
```diff
@@ -0,0 +1,283 @@
...
+
+    // Calculated values
+    private DataFrame output;
+
+    public MultiMADClassifier(List<String> columnNames) {
+        this.columnNames = columnNames;
+    }
+
+    public MultiMADClassifier(String... columnNames) {
+        this.columnNames = new ArrayList<String>(Arrays.asList(columnNames));
```
sergeykhegay: `new ArrayList<String>(..)` is redundant here. `Arrays.asList(..)` already returns a newly allocated `ArrayList`.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/sample/Sampler.java
```diff
@@ -0,0 +1,36 @@
...
+    public double[] getSample(double[] input) {
+        int sampleSize = sampleIndices.length;
+        double[] samples = new double[sampleSize];
+        for (int i = 0; i < sampleSize; i++) {
+            samples[i] = input[sampleIndices[i]];
+        }
+        return samples;
+    }
+
+    public String[] getSample(String[] input) {
```
sergeykhegay: I would suggest using generics here. Reduces codebase and might help preventing future mistakes when someone edits one method and forgets to edit the other.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/sample/ReservoirSampler.java
```diff
@@ -0,0 +1,38 @@
+package edu.stanford.futuredata.macrobase.analysis.sample;
+
+import java.util.Random;
+
+/**
+ * Samples using reservoir sampling
+ */
+public class ReservoirSampler extends Sampler {
+    private Random rand;
```
sergeykhegay: This could have gone to your abstract class

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/sample/FisherYatesSampler.java
```diff
@@ -0,0 +1,40 @@
+package edu.stanford.futuredata.macrobase.analysis.sample;
+
+import java.util.Random;
+
+/**
+ * Samples using a partial Fisher-Yates shuffle: an array of the population size is shuffled
+ * until a sample of the specified size is obtained.
+ */
+public class FisherYatesSampler extends Sampler {
+    private Random rand;
```
sergeykhegay: This could have gone to your abstract class.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/sample/Sampler.java
```diff
@@ -0,0 +1,36 @@
+package edu.stanford.futuredata.macrobase.analysis.sample;
+
+/**
+ * Abstract class for sampling. Should not modify input.
+ * computeSampleIndices computes and stores the indices of samples.
+ * getSamplingMethod returns the type of the sampler.
+ * getSamples returns an array of samples from the input using the computed indices.
+ */
+public abstract class Sampler {
```
sergeykhegay: Is there a specific reason for this kind of interface? Is it intended to pre-generate sampling indices once and then call `getSample(...)` on the same indices for different data sets?

If not, then you might consider marking something like this:
```
    // Returns an array with sample indices
    abstract int[] computeSampleIndices(int populationSize, double samplingRate);
   
    public double[] getSample(double[] input, double sampleRate) {
        // check inputs here
        int[] sampleIndices = computeSampleIndices(input.size(), sampleRate);
        .....
    }
    
     or with generics

    public <T> T[] getSample(T[] input, double sampleRate) {
        // check inputs here
        int[] sampleIndices = computeSampleIndices(input.size(), sampleRate);
        .....
    }
```

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/sample/FisherYatesSampler.java
```diff
@@ -0,0 +1,40 @@
...
+            int j = rand.nextInt(populationSize - i) + i;
+            int temp = range[j];
+            range[j] = range[i];
+            range[i] = temp;
+        }
+        System.arraycopy(range, 0, sampleIndices, 0, sampleSize);
+    }
+
+    public String getSamplingMethod() {
+        return "fisher-yates";
```
sergeykhegay: There should be no magic constants. Might be better to declare 
```
private static final String SAMPLING_METHOD = "fisher-yates";
....

public String getSamplingMethod() {
        return SAMPLING_METHOD;
}
```

PR #201, opened 2017-06-23
initial version of cli package

edgan8:
Preliminary cli interface for hooking up operators from lib into pipelines configured by yaml files.

Basically a stripped down / cleaned up version of the old command line flow but now using the lib operators. 

I found this very useful to try out new ideas on new data that require more customization than the ui.


edgan8:
cc @kexinrong @pbailis

cli/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BatchPipeline.java
```diff
@@ -0,0 +1,68 @@
...
+        loader.setColumnTypes(colTypes);
+        DataFrame df = loader.load();
+        System.out.println(df.getNumRows());
+        System.out.println(attributes);
+
+        PercentileClassifier classifier = new PercentileClassifier(metric);
+        classifier.setPercentile(percentile);
+        classifier.setIncludeHigh(includeHi);
+        classifier.setIncludeLow(includeLo);
+        classifier.process(df);
```
kexinrong: I think we can chain these setters together?

> classifier.setPercentile().setIncludeHigh()...

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/PercentileClassifier.java
```diff
@@ -9,21 +9,19 @@
...
     private double percentile = 0.5;
     private boolean includeHigh = true;
     private boolean includeLow = true;
-    private String columnName;
-    private String outputColumnName = "_OUTLIER";
 
     // Calculated values
     private double lowCutoff;
     private double highCutoff;
     private DataFrame output;
```
kexinrong: This dataframe output variable should also be in the Classifier class?

cli/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BatchPipeline.java
kexinrong: Sorry didn't realize some of the setters return Classifer class instead

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/PercentileClassifier.java
edgan8: Yeah, each classifier should save the column name it outputs results in so I've moved it into the base class. getResults will change for each implementation though: some classifiers might cache it and some might want to compute stuff on demand.

PR #203, opened 2017-07-13
Fixed compatibility with AvaticaResultSet which doesn't support isLast()

jarnoux:
Hello,
This PR fixes compatibility with the avatica JDBC driver, which doesn't support `isLast()` [1].

Tests are passing, please let me know if anything else needs to be done to merge.

Jacques

[1] https://github.com/apache/calcite-avatica/blob/4db1fb9c66db8ccebc9e96ce678154ec69c557f0/core/src/main/java/org/apache/calcite/avatica/AvaticaResultSet.java#L444


jarnoux:
Is the coverage going to be an issue? I could artificially restore the line count to make it pass but that's not very useful

pbailis:
LGTM; thanks for this!

PR #205, opened 2017-07-15
REST Server for lib functionality

pbailis:


pbailis:
@edgan8 good to go

pbailis:
After conversations with @fabuzaid21, we decided to put both a generic `Pipeline` class and the REST server under a new module called `core`. Therefore, `cli` depends on `core` (and only `core`, not `lib`), and new pipelines can be put into `core`, too, without affecting lib.

Alternatives we discussed closely:
 * Put pipelines in `lib` (problem: since `lib` is our "stable" makes it very high bar to add more pipelines)
 * Duplicate pipeline functionality in `cli` and `server` (problem: if we add new pipelines, they'll be incompatible across CLI and the REST API)
 * Create a separate `pipeline` package (problem: seems overkill in terms of resulting dependencies -- `server` -> `pipeline` -> `lib`)

Apache Spark also uses the `server`-equivalent in its `core`, and the new Spark web framework makes it much more lightweight.

edgan8:
Having 3 levels of dependencies could get clunky to manage. I think either or both of

a) merging cli and core into one package (core) 
b) putting pipelines in lib and commenting that they are unstable until ready

would in practice be way more convenient

pbailis:
Following the discussion with @sahaana and @fabuzaid21, confirming we will put pipelines in `core` along with `cli`, with experimental modules also in `core` but clearly delineated as such. @edgan8 do you feel comfortable finishing this refactoring?

pbailis:
I think it'd be awesome to have an annotation like @Experimental as in RxJava (http://reactivex.io/RxJava/javadoc/rx/annotations/Experimental.html) or Apache Spark (https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/annotation/Experimental.html)

edgan8:
Sounds good. An annotation would work great, and I am also happy with annotating the class javadocs for now.

edgan8:
I'm currently updating this to fold cli into core. Also taking this chance to unify the configuration / query objects for the cli and server pipelines.

edgan8:
@fabuzaid21 @jialinding curious what you guys think about this refactoring

PR #208, opened 2017-07-20
Add pipeline elements for cubed data

jialinding:


jialinding:
Still need to rebase on the REST server code.

lib/src/test/resources/sample_cubed.csv
```diff
@@ -0,0 +1,10 @@
+location,version,count,mean,std
```
edgan8: Also add a .yaml demo file so people can try running the cli on cubed data.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/ArithmeticClassifier.java
```diff
@@ -0,0 +1,162 @@
...
+        this.stdColumnName = stdColumnName;
+    }
+
+    @Override
+    public void process(DataFrame input) {
+        double[] means = input.getDoubleColumnByName(meanColumnName);
+        double[] counts = input.getDoubleColumnByName(countColumnName);
+        double[] stds = input.getDoubleColumnByName(stdColumnName);
+        int len = means.length;
+        int numRawMetrics = 0;
```
jialinding: hmm but then it's awkward when I use these variables to set the length of arrays and I have to cast them back to ints

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/ArithmeticClassifier.java
```diff
@@ -0,0 +1,162 @@
...
+        highCutoff = new Percentile().evaluate(rawMetrics, 100.0 - percentile);
+
+        output = input.copy();
+        double[] resultColumn = new double[len];
+        for (int i = 0; i < len; i++) {
+            double mean = means[i];
+            double std = stds[i];
+            double count = counts[i];
+            double numOutliers = 0.0;
+            if (Double.isNaN(std) || std == 0.0) {
```
jialinding: if there's only one raw event in that group, the standard deviation comes out to NaN, at least when I did aggregation with Pandas.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/ArithmeticClassifier.java
```diff
@@ -0,0 +1,162 @@
+package edu.stanford.futuredata.macrobase.analysis.classify;
+
+import edu.stanford.futuredata.macrobase.datamodel.DataFrame;
+
+import java.lang.Double;
```
jialinding: I'm using Double for the isNaN check

lib/src/main/java/edu/stanford/futuredata/macrobase/ingest/CSVDataFrameLoader.java
```diff
@@ -57,23 +57,23 @@ public DataFrame load() throws Exception {
...
+                String rowValue = record.get(c);
+                if (t == Schema.ColType.STRING) {
+                    rowFields.add(rowValue);
+                } else if (t == Schema.ColType.DOUBLE) {
+                    try {
                         rowFields.add(Double.parseDouble(rowValue));
-                    } else {
-                        throw new RuntimeException("Bad ColType");
+                    } catch (NumberFormatException e) {
+                        rowFields.add(Double.NaN);
```
jialinding: Yeah, when a certain value in a double column was NaN (like standard deviation), the row would be thrown out

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/APrioriSummarizer.java
```diff
@@ -50,16 +56,19 @@ public void process(DataFrame input) throws Exception {
...
+            hasCountCol = true;
+        }
+        numTotal = 0;
+        double numOutliersExact = 0.0;
         for (int i = 0; i < n; i++) {
-            flag[i] = predicate.test(outlierCol[i]);
-            if (flag[i]) {
-                numOutliers++;
-            }
+            numTotal += hasCountCol ? countCol[i] : 1;
```
jialinding: cool, I changed this one, but there are two other similar pieces of code in countSingles and countSet where it's harder to move the condition outside the loop without adding a lot of redundant code, so I left those unchanged.

cli/bin/simple.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 BIN=`dirname "$0"`
 BASE=$BIN/..
-java -Xmx10g -cp "$BASE/config:$BASE/target/classes:$BASE/target/*" \
-edu.stanford.futuredata.macrobase.runner.SimpleRunner "$@"
+java -Xmx10g -cp "$BASE/config:$BASE/target/classes:$BASE/../lib/target/classes:$BASE/target/*" \
```
jialinding: This probably isn't the right way to do things, but for some reason the NormalDist LUT resource file in lib was causing a FileNotFoundException when I ran the pipeline from CLI, so I manually added it to the classpath here.

lib/src/test/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/NormalDistTest.java
```diff
@@ -0,0 +1,61 @@
...
+        cdf = dist.cdf(0.0, 1.0, 4.0);
+        assertEquals(1.0, cdf, 0.001);
+
+        // Test min z-score
+        cdf = dist.cdf(0.0, 1.0, -4.0);
+        assertEquals(0.0, cdf, 0.001);
+    }
+
+    @Test
+    public void constructLUT() {
```
jialinding: Not sure if I should comment this out, or remove it altogether.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private void computeCounts() {
+        metricCounts = HashMultiset.create();
+        int len = counts.length;
+        for (int i = 0; i < len; i++) {
+            metricCounts.add(metrics[i], (int) counts[i]);
+        }
+        numRawMetrics = metricCounts.size();
+        sortedMetrics = new ArrayList<Double>(metricCounts.elementSet());
+        Collections.sort(sortedMetrics);
```
edgan8: Arrays.sort and using raw arrays should be faster

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Multiset;
+
+import java.util.*;
+
+/**
+ * Computes percentiles when given an array of metrics and the count of the
+ * number of times each occurs. Useful for computing percentiles on cubed data.
+ */
+public class CubePercentile {
```
edgan8: Rename this to WeightedPercentile which is its more general use

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
```diff
@@ -0,0 +1,60 @@
...
+            int numToPass = (int)(percentile / 100.0 * numRawMetrics);
+            int numPassed = 0;
+            for (int i = 0; i < sortedMetrics.size(); i++) {
+                numPassed += metricCounts.count(sortedMetrics.get(i));
+                if (numPassed >= numToPass) {
+                    return sortedMetrics.get(i);
+                }
+            }
+        }
+        throw new RuntimeException("CubePercentile was implemented incorrectly");
```
edgan8: throw MacrobaseInternalError

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
```diff
@@ -0,0 +1,60 @@
...
+        this.metrics = metrics;
+        computeCounts();
+    }
+
+    public double evaluate(double percentile) {
+        if (percentile >= 50.0) {
+            int numToPass = (int)((100.0 - percentile) / 100.0 * numRawMetrics);
+            int numPassed = 0;
+            for (int i = sortedMetrics.size() - 1; i >= 0; i--) {
+                numPassed += metricCounts.count(sortedMetrics.get(i));
```
edgan8: If you precalculated and cached of the running sum of the counts you probably wouldn't need to add in the loop. Could just do straight comparisons instead or even do binary search.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
```diff
@@ -0,0 +1,60 @@
...
+ * Computes percentiles when given an array of metrics and the count of the
+ * number of times each occurs. Useful for computing percentiles on cubed data.
+ */
+public class CubePercentile {
+    private double[] counts;
+    private double[] metrics;
+
+    // Computed
+    private int numRawMetrics = 0;
+    private Multiset<Double> metricCounts;
```
edgan8: Multiset will be pretty slow when you have groups with large counts: remember the big advantage of using cubes is to minimize memory usage and time by not having the explode out the raw rows.

Instead of metricCounts and sortedMetrics we should try something like WeightedMetric[] metrics. where WeightMetric is a metric value and a count/weight.

To see this in action try creating a unit test where you have cubes with only 2 dimensions and say counts of 1Million: this code will be relatively slow in that case when it could be way faster if you don't explode out the counts.

lib/src/main/java/edu/stanford/futuredata/macrobase/ingest/CSVDataFrameLoader.java
edgan8: good catch

lib/src/test/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/NormalDistTest.java
edgan8: I would precompute it once and then just save the array statically. You can have a function in the WeightedPercentile class then that recomputes the array so you can compare the static value with the computed one in a unit test.

edgan8:
This code is looking better! My most important feedback is to have a test where you make sure the code *scales* to large counts which is the killer use case of cubes in the first place, if you do you should run into some of the issues I outlined.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
jialinding: I think Multiset doesn't actually store duplicate elements. It seems to me like a version of Map that's good at counting, which is exactly what I want here, so it doesn't store much more than an array.

I don't think arrays are good here, because we don't know the number of unique metrics ahead of time, and we can't efficiently find and update elements with counts. Lists also run into the problem of finding and updating. So we want a map-like data structure.

I coded up a version of WeightedMetrics[] and benchmarked it. It's much slower than the current implementation.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
jialinding: I benchmarked, and it seems like pre-calculating the running sum takes much more time than the amount of time that's saved in evaluate(). This is especially true given our use case, where we only call evaluate() twice.

Indeed, the vast majority of time is spent constructing the relevant lists and arrays, and not much time is spent evaluating percentiles (maybe a 20 to 1 ratio).

Also in our use case, straight comparisons is better than binary search because we're usually evaluating very high and very low percentiles anyway.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
jialinding: Makes sense.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/NormalDist.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Implements a look-up table in order to quickly obtain the CDF of a
+ * normal distribution.
+ */
+public class NormalDist {
+    public static final double GRANULARITY = 0.01;
+    public static final double MAXZSCORE = 3.5;
+    public static final double MINZSCORE = -3.5;
+
+    private Map<Integer, Double> cdfLUT;
```
jialinding: Yep, using a double[] comes out 250x faster than the standard library cdf :)

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/CubePercentile.java
edgan8: Cool makes sense. This kind of benchmarking is the key to fast code!

cli/src/main/java/edu/stanford/futuredata/macrobase/pipeline/CubePipeline.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public void run() throws Exception {
+        Map<String, Schema.ColType> colTypes = getColTypes();
+        CSVDataFrameLoader loader = new CSVDataFrameLoader(inputFile);
+        loader.setColumnTypes(colTypes);
+        long startTime = System.currentTimeMillis();
+        DataFrame df = loader.load();
+        long elapsed = System.currentTimeMillis() - startTime;
+        log.info("Loading time: {}", elapsed);
```
edgan8: Use debug for diagnostic data, info for things normal end users might actually care about.

cli/src/main/java/edu/stanford/futuredata/macrobase/pipeline/CubePipeline.java
```diff
@@ -0,0 +1,122 @@
...
+            case "arithmetic": {
+                ArithmeticClassifier classifier =
+                        new ArithmeticClassifier(countColumn, meanColumn, stdColumn);
+                classifier.setPercentile(percentile);
+                classifier.setIncludeHigh(includeHi);
+                classifier.setIncludeLow(includeLo);
+                return classifier;
+            }
+            default:
+                return null;
```
edgan8: throw an exception here for better error messages

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/DiscreteClassifier.java
```diff
@@ -0,0 +1,115 @@
...
+        highCutoff = wp.evaluate(100.0 - percentile);
+
+        output = input.copy();
+        double[] resultColumn = new double[len];
+        for (int i = 0; i < len; i++) {
+            double mean = means[i];
+            if ((mean > highCutoff && includeHigh)
+                    || (mean < lowCutoff && includeLow)
+                    ) {
+                resultColumn[i] = 1.0;
```
edgan8: shouldn't this be counts[i]?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/DiscreteClassifier.java
```diff
@@ -0,0 +1,115 @@
...
+
+import edu.stanford.futuredata.macrobase.analysis.classify.stats.WeightedPercentile;
+import edu.stanford.futuredata.macrobase.datamodel.DataFrame;
+
+/**
+ * Classify rows by high / low values based on the group mean.
+ * Returns a new dataframe with a column representation of the estimated number of outliers
+ * for each group, which will either be zero or all elements in the group.
+ */
+public class DiscreteClassifier extends CubeClassifier implements ThresholdClassifier {
```
edgan8: No test and it isn't used anywhere, might be worth deleting this or adding test and actually making it possible to use from CubeClassifier

PR #210, opened 2017-08-05
Parsing json results of cube service

edgan8:
Parsing the json outputs of the cube service response into dataframes.

aria_api_cube contains generic data in the same shape as we will expect from the cube service


core/src/main/java/edu/stanford/futuredata/macrobase/contrib/aria/APICubeResult.java
```diff
@@ -0,0 +1,10 @@
+package edu.stanford.futuredata.macrobase.contrib.aria;
```
pbailis: kind of strange this lives in `core` even though it's really `contrib`... do we want to rethink here, or just let core contain basically everything

core/src/main/java/edu/stanford/futuredata/macrobase/contrib/aria/APICubeResult.java
```diff
@@ -0,0 +1,10 @@
...
+
+import java.util.List;
+
+public class APICubeResult {
+    public String interval;
+    public String granularity;
```
pbailis: comments here to describe what each field means would be really useful; e.g., what is interval vs granularity?

core/src/main/java/edu/stanford/futuredata/macrobase/contrib/aria/APIFilterResult.java
```diff
@@ -0,0 +1,13 @@
+package edu.stanford.futuredata.macrobase.contrib.aria;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class APIFilterResult {
+    public String dimension;
+    public String operation;
```
pbailis: Ditto, comments here would be really helpful.

core/src/main/java/edu/stanford/futuredata/macrobase/contrib/aria/APISeriesResult.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class APISeriesResult {
+    public String operation;
+    public List<Double> values;
+    public String segment;
```
pbailis: Ditto, comment, e.g., what is 'segment'?

edgan8:
No longer needed since we have removed most cube specific code!

PR #211, opened 2017-08-06
First commit of RawThresholdClassifier

fabuzaid21:


core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
```diff
@@ -2,17 +2,19 @@
 
 import edu.stanford.futuredata.macrobase.analysis.classify.Classifier;
 import edu.stanford.futuredata.macrobase.analysis.classify.PercentileClassifier;
+import edu.stanford.futuredata.macrobase.analysis.classify.RawThresholdClassifier;
```
pbailis: Nit:  any reason why not to do `ThresholdClassifier`?

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
```diff
@@ -21,44 +23,54 @@
...
         summarizerType = conf.get("summarizer");
         attributes = conf.get("attributes");
         minRiskRatio = conf.get("minRiskRatio");
         minSupport = conf.get("minSupport");
-
     }
 
     public Classifier getClassifier() throws MacrobaseException {
         switch (classifierType.toLowerCase()) {
             case "percentile": {
```
pbailis: Use ENUM here with str.toUpper?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/RawThresholdClassifier.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @return Lambda function corresponding to the ``predicateStr''. The Lambda function takes in a single
+     * argument, which will correspond to the value in the metric column. (A closure is created around the ``sentinel''
+     * parameter.)
+     * @throws MacrobaseException
+     */
+    private DoublePredicate getPredicate(final String predicateStr, final double sentinel) throws MacrobaseException {
+        switch (predicateStr) {
+            case "==":
```
pbailis: want to trim predicateStr?

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
fabuzaid21: There's already a `ThresholdClassifier` in `lib/`, the exact same package

fabuzaid21: It's an interface that `ArithmeticClassifier` implements

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
fabuzaid21: agree with Enum suggestion—but I was thinking it was more applicable to the predicates, since there's 6 predicates that we currently support. How many classifiers do you envision we'll have?

pbailis: fine to leave as str

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
pbailis: Minor nit, but I would rename ArithmeticClassifier to something like StdDevGroupClassifier

PR #214, opened 2017-08-28
Add QuantileClassifier

jialinding:


lib/src/test/java/edu/stanford/futuredata/macrobase/analysis/classify/QuantileClassifierTest.java
```diff
@@ -0,0 +1,120 @@
...
+        double trueLowCutoff = percentile.evaluate(1);
+        double trueHighCutoff = percentile.evaluate(99);
+        assertEquals(trueLowCutoff, ac.getLowCutoff(), 5.0);
+        assertEquals(trueHighCutoff, ac.getHighCutoff(), 5.0);
+
+        double[] outliers = output.getDoubleColumnByName("_OUTLIER");
+
+        for (int i = 0; i < outliers.length; i++) {
+            int trueNumOutliers = 0;
+            trueNumOutliers += Math.max(0, trueLowCutoff - (i-50));
```
edgan8: This math gets very confusing: if we just save rawdata it's really easy to count how many are actually above / below the threshold. That also makes the test more robust to when you change the data generation logic.

jialinding: Yeah, I just wasn't sure if we wanted to have to save a List of arrays, in order to save all the raw data for each group. That would be in addition to the array of raw data that I'm already saving. Or I could just save the List of arrays and reconstruct the array of raw data by unrolling the List of arrays, but that seemed like a bunch of unnecessary logic and copying.

edgan8:
pushit!

PR #216, opened 2017-08-29
generalized linear apriori pruning

edgan8:
support multiple custom "risk ratio-like", "support-like", AND "regression" cube scenarios all in one algorithm. This new structure also gives more intuition for the essential features of a-priori + additional pruning: which is that the additional pruning must be based on FUNCTIONS of LINEAR AGGREGATES that may or may not be monotonic.

See APLOutlierSummary to see how the classical risk ratio + support style usage is implemented in this abstract framework.


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLOutlierSummarizer.java
```diff
@@ -0,0 +1,91 @@
...
+        );
+        long elapsed = System.currentTimeMillis() - startTime;
+        log.debug("Encoded in: {}", elapsed);
+        log.debug("Encoded Categories: {}", encoder.getNextKey());
+
+        double[][] aggregateColumns = new double[2][];
+        aggregateColumns[0] = outlierCol;
+        aggregateColumns[1] = countCol;
+        aplResults = aplKernel.explain(encoded, aggregateColumns);
+        System.out.println(aplResults);
```
jialinding: Should the results be sorted before printing?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanationResult.java
```diff
@@ -0,0 +1,26 @@
...
+            double[] metrics
+    ) {
+        this.matcher = matcher;
+        this.aggregates = aggregates;
+        this.metrics = metrics;
+    }
+
+    @Override
+    public String toString() {
+        return "a="+Arrays.toString(matcher.values)+":ag="+Arrays.toString(aggregates)+":mt="+Arrays.toString(metrics);
```
jialinding: Maybe put the name of the quality metric in the output, so users know what the metric they're seeing means.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanationResult.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Arrays;
+
+public class APLExplanationResult {
+    private IntSet matcher;
+    private double[] aggregates;
+    private double[] metrics;
+
+    public APLExplanationResult(
+            IntSet matcher,
```
jialinding: Why not put in the decoded IntSet, like in ExplanationResult?

edgan8:
@fabuzaid21 @pbailis This is code to generalize apriori mining to support 1) row-based outlier ratio explanation 2) group-by based explanation 3) mean-shift "regression" explanation all with the same core APLinear routine. See for example how APLMeanSummarizer and APLOutlierSummarizer can be implemented very simply by customizing the generic routine a little.

Microsoft wants to try the mean-shift explanation so I want to get it merged in and then hooked up to the aria pipeline. 

https://futuredata.slack.com/files/U0MFTDX8S/F6VC9FCAF/Generalized_Apriori_Pruning for the high level intuition behind how this generalization works.

PR #220, opened 2017-11-21
Incremenatal Summarizer

kexinrong:
- Make Incremental Summarizer compatible with master API
- Prune by risk ratio (in addition to support) when tracking itemsets


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/fpg/IncrementalSummarizer.java
```diff
@@ -288,16 +323,22 @@ private void addNewFrequent() {
         for (ItemsetWithCount iwc : frequent) {
             Set<Integer> itemset = iwc.getItems();
             if (!outlierItemsetWindowCount.containsKey(itemset)) {
-                trackItemset(itemset, iwc.getCount());
                 newFrequent.add(iwc);
             }
         }
         // Get support in the inlier transactions
         List<ItemsetWithCount> inlierCount = fpGrowth.getCounts(
```
edgan8: let's rename this to newFrequentInlierCounts to be more explicit about exactly what it's tracking

PR #221, opened 2017-11-21
MCD train; CSV ingester row count

kexinrong:
- Pass in a copy of input data to MCD so that the order of the original input is maintained
- Fix CSV ingester total row count


legacy/src/main/java/macrobase/ingest/CSVIngester.java
```diff
@@ -103,7 +103,7 @@ private Datum parseRecord(CSVRecord record) throws NumberFormatException {
                     badRows++;
                 }
             }
-            log.info("{}/{} rows successfully parsed ({} malformed rows)", numRows-badRows, numRows, badRows);
```
edgan8: lol nice catch. Was this actually causing issues?

legacy/src/main/java/macrobase/analysis/transform/BatchScoreFeatureTransform.java
```diff
@@ -39,8 +40,11 @@ public void initialize() {
 
     @Override
     public void consume(List<Datum> records) {
-        if(requiresTraining)
-            batchTrainScore.train(records);
+        if(requiresTraining) {
+            List<Datum> copy = new ArrayList<>(records);
```
edgan8: Not every featuretransform is going to modify its records. Can we add the copy to MCD so other transforms don't need to pay the overhead?

legacy/src/main/java/macrobase/ingest/CSVIngester.java
kexinrong: Not really. I just noticed that the line count numbers were wrong.

legacy/src/main/java/macrobase/analysis/transform/BatchScoreFeatureTransform.java
kexinrong: Make sense. Will do.

PR #227, opened 2017-11-22
Adding predicate query support and prevalence ratio

sahaana:


core/demo/cli_cube_predicate.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+inputURI: "csv://core/demo/sample_cubed.csv"
+
+classifier: "predicate"
+countColumn: "count"
+meanColumn: "mean"
+stdColumn: "std"
+
+metric: "mean"
+predicate: "<="
```
edgan8: Lol it's hard to tell if the output for this query actually corresponds to "correct" output right? Can we create a new csv file that realy showcases the predicate classifier?

core/demo/cli_predicate.yaml
```diff
@@ -0,0 +1,15 @@
+pipeline: "BasicBatchPipeline"
+
+inputURI: "csv://core/demo/tiny_predicate.csv"
+
+classifier: "predicate"
+metric: "usage"
+predicate: "=="
```
edgan8: Same, checking for floating point equailty on actual floats is kind of weird, maybe turn this into a slightly more reailstic example? That way we can actually refer to the demos as exemplary documentation of macrobaes best practice.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/PredicateCubeClassifier.java
```diff
@@ -0,0 +1,160 @@
+package edu.stanford.futuredata.macrobase.analysis.classify;
+
+import edu.stanford.futuredata.macrobase.datamodel.DataFrame;
+import edu.stanford.futuredata.macrobase.util.MacrobaseException;
+
+import java.util.function.DoublePredicate;
```
edgan8: confused, why does this file show up twice?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/ratios/PrevalenceRatioMetric.java
```diff
@@ -0,0 +1,31 @@
+package edu.stanford.futuredata.macrobase.analysis.summary.ratios;
+
+/**
+ * P(exposure | outlier) / P(exposure | ~outlier)
+ */
+public class PrevalenceRatioMetric extends ExplanationMetric {
```
edgan8: does this need to be part of this PR?

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
```diff
@@ -24,45 +26,48 @@
...
-    private boolean pctileLow;
-
-    private String summarizerType;
-    private List<String> attributes;
-    private String ratioMetric;
-    private double minSupport;
-    private double minRiskRatio;
+    // PipelineConfig params applicable to all classifiers
+    private final String metric;
+    private final String inputURI;
```
edgan8: I agree it's a good idea to move commonly used config params to a common place, however one thing I don't like about this approach is that now all of the conf params a pipeline needs are spread out in random places. I would much rather have them gathered in a way where it's easy to find all of the params one might need for a given pipeline.

What do we really gain from moving the conf parameter lookups to the clasifier / summarizer initialization section? Can we get those benefits while still making the complete set of parameters easy to find?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/PredicateCubeClassifier.java
sahaana: This code duplication is a ugly hack that makes the predicate classifier work with cube pipelines (since cube classifiers and classifiers are separate). An alternative proposal (thanks @kexinrong) would be to create a classifier class, and have all other classifiers (cube and not) just extend this. The cube classifiers would all need to add their own count columns. This might be ok since we only check for a certain number of classifiers in the cube/batch pipelines. We should discuss which works best once you're back!

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/ratios/PrevalenceRatioMetric.java
sahaana: I can split this into a separate PR

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/CubePipeline.java
```diff
@@ -157,6 +158,14 @@ private CubeClassifier getClassifier() throws MacrobaseException {
                 classifier.setIncludeLow(includeLo);
                 return classifier;
             }
+            case "predicate": {
+                // default values for PredicateClassifier:
+                // {predicate: "==", value: 1.0}
+                final String metric = conf.get("metric");
+                final String predicateStr = conf.get("predicate", "==").trim();
```
edgan8: Same here I think it's easier to track if we do all conf loading up front in one spot and then only use the loaded instance variables in the body of the class

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
```diff
@@ -28,9 +29,13 @@
 
     private String classifierType;
     private String metric;
+    private Object rawCutoff;
```
edgan8: this can just be a local variable

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/BasicBatchPipeline.java
```diff
@@ -113,6 +139,9 @@ public BatchSummarizer getSummarizer(String outlierColumnName) throws MacrobaseE
     public DataFrame loadData() throws Exception {
         Map<String, Schema.ColType> colTypes = new HashMap<>();
         colTypes.put(metric, Schema.ColType.DOUBLE);
```
edgan8: make this into if/else

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/CubePipeline.java
```diff
@@ -29,41 +24,69 @@
 public class CubePipeline implements Pipeline {
     Logger log = LoggerFactory.getLogger("CubePipeline");
 
+    private final PipelineConfig conf;
```
edgan8: cut

core/src/main/java/edu/stanford/futuredata/macrobase/pipeline/CubePipeline.java
```diff
@@ -29,41 +24,69 @@
...
     private Map<String, Object> jsonBody;
     private boolean usePost;
 
+    // Classifiers
     private String classifierType;
     private String countColumn;
-    private String meanColumn;
-    private String stdColumn;
-    private LinkedHashMap<String, Double> quantileColumns;
+    private Object rawCutoff;
```
edgan8: can be local var

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/classify/stats/MBPredicate.java
```diff
@@ -0,0 +1,94 @@
...
+            case EQUALS:
+                return (String x) -> x.equals(sentinel);
+            case NOT_EQUALS:
+                return (String x) -> !(x.equals(sentinel));
+            case LESS_THAN:
+                return (String x) -> x.compareTo(sentinel) < 0;
+            case GREATER_THAN:
+                return (String x) -> x.compareTo(sentinel) > 0;
+            case LEQ:
+                return (String x) -> (x.equals(sentinel) || x.compareTo(sentinel) < 0);
```
edgan8: for strings can do compareTo <= 0

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/BatchSummarizer.java
```diff
@@ -18,6 +18,8 @@
     // Parameters
     protected String outlierColumn = "_OUTLIER";
     protected double minOutlierSupport = 0.1;
+    protected double minRatioMetric = 3;
+    protected int maxOrder;
```
edgan8: not used anywhere

PR #229, opened 2017-12-01
Add host command line parameter in load_demo.py, use the same command…

rockie-yang:
Add host parameter to load_demo.py, so that the postgres database does not have to run on the same machine.

It enables create test environment with docker-compose, postgres runs in a container and macrobase runs in another containter.


tools/load_demo2.py
```diff
@@ -12,6 +12,17 @@
 if(len(sys.argv) > 2):
     port = sys.argv[2]
 
+user = "postgres"
+password = "mamikonyan"
```
kexinrong: Can we change the default password to something like 'postgres' or 'admin'? Thanks!

kexinrong:
Thanks for the pull request!

PR #232, opened 2017-12-09
calculate total outliers

edgan8:
as requested

`curl -H "Accept: application/json" -H "Content-type: application/json" -X POST -d "@core/demo/aria_meanshift.json" 0.0.0.0:4567/query; echo`

```
{"numTotal":2.7315584E7,"outliers":213823.0,"results":[{"metric":{"global_ratio":"37.678","support":"0.982"},"matcher":{"Startup Type":"ColdStartUp"},"aggregate":{"Outliers":"209920.944","Count":"711744.000"}},{"metric":{"global_ratio":"2.274","support":"0.013"},"matcher":{"OS Version":"4.0"},"aggregate":{"Outliers":"2800.818","Count":"157366.000"}}]}
```

### DIFF
--- a/lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanation.java
+++ b/lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanation.java
@@ -13,6 +13,7 @@ public class APLExplanation implements Explanation {
     private AttributeEncoder encoder;
     private List<String> aggregateNames;
     private long numTotal;
+    private long numOutliers;
 
     private ArrayList<QualityMetric> metrics;
     private List<Double> thresholds;
@@ -21,6 +22,7 @@ public class APLExplanation implements Explanation {
     public APLExplanation(
             AttributeEncoder encoder,
             long numTotal,
+            long numOutliers,
             List<String> aggregateNames,
             List<QualityMetric> metrics,
             List<Double> thresholds,
@@ -28,6 +30,7 @@ public class APLExplanation implements Explanation {
     ) {
         this.encoder = encoder;
         this.numTotal = numTotal;
+        this.numOutliers = numOutliers;
         this.aggregateNames = aggregateNames;
         this.metrics = new ArrayList<>(metrics);
         this.thresholds = new ArrayList<>(thresholds);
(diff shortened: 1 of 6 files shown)


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanation.java
```diff
@@ -52,11 +55,17 @@ public double numTotal() {
...
+    public double numOutliers() {
+        return numOutliers;
+    }
+
     @Override
     public String prettyPrint() {
         StringBuilder header = new StringBuilder(String.format(
                 "Outlier Explanation:\n"
         ));
+        header.append("Outliers: "+numOutliers+", Total: "+numTotal+"\n");
```
fabuzaid21: This is supposed to print the JSON response, right? It doesn't exactly match the example you put in the commit message—is that okay? (Not sure how strict they are about the output.)

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLMeanSummarizer.java
```diff
@@ -70,6 +70,25 @@
         return Arrays.asList(minSupport, minStdDev);
     }
 
+    @Override
+    public double getNumberOutliers(double[][] aggregates) {
+        int n = aggregates[0].length;
+        int k = aggregates.length;
```
fabuzaid21: Will aggregates.length ever equal 0? If so, should we initialize `k` before `n` and check? (Sorry, not super-familiar with this code.)

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanation.java
edgan8: no this is the pretty print, the json response is defined by the @Jsonfield decorator

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLMeanSummarizer.java
edgan8: In this case aggregates is created by the getAggregates function in the class so this class controls the size. It will always have 3 columns.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanation.java
fabuzaid21: Oh, I see—nvm!

PR #235, opened 2018-01-11
Get aplinear working in BasicBatchPipeline, improving interfaces in the process

kraftp:


edgan8:
OK I tried this out, I changed to "summarizer": "aplinear" in the batch.yaml and batch.json and it seems to work well =p. Please update those configs to use aplinear in this PR and let's shipit. Thanks for updating the code!

PR #240, opened 2018-01-19
Generalized FPGrowth

kraftp:


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/fpg/FPGrowthEmerging.java
```diff
@@ -147,33 +189,48 @@ public FPGrowthEmerging setCombinationsEnabled(boolean flag) {
...
+                }
+            }
 
-            if (ratio >= minRatio) {
-                ret.add(new FPGItemsetResult(oc.getCount() / (double) outliers.size(),
-                                          oc.getCount(),
-                                          ratio,
-                                          oc.getItems()));
+            if (passThresholds) {
+                double riskRatio = RiskRatio.compute(ic.getCount(),
```
edgan8: Why do we hardcode RiskRatio here? Since we're already generalizing fpgrowth to suport arbitrary quality metrics (with support still hardcoded) we should use the user-provided qualitymetric.

PR #241, opened 2018-01-24
Don't run tests in build.sh

pbailis:


edgan8:
What's the motivation here?

edgan8:
actually I kind of see the motivation. move fast.

build.sh
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-cd lib && mvn clean && mvn install &&  cd ../sql && mvn clean && mvn package
+cd lib && mvn clean && mvn install &&  cd ../sql && mvn clean && mvn package -DskipTests
```
edgan8: Firas why did you remove core from the global build script?

PR #244, opened 2018-02-01
Modular optimized apriori

kraftp:
Now fully supports Order-3 analyses on high-cardinality data!  Also has more modular data structures and implements a cascade system.


lib/src/test/java/edu/stanford/futuredata/macrobase/analysis/summary/apriori/APrioriSummarizerTest.java
```diff
@@ -1,155 +0,0 @@
...
-import org.junit.Test;
-
-import java.util.*;
-
-import static junit.framework.TestCase.assertEquals;
-import static org.junit.Assert.assertTrue;
-
-public class APrioriSummarizerTest {
-    @Test
-    public void testSimple() throws Exception {
```
fabuzaid21: @kraftp We need to keep this test around—we still need tests of the BasicBatchPipeline.

kraftp: Done

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLExplanationResult.java
```diff
@@ -55,11 +56,6 @@ public APLExplanationResult(
         return aggregate;
     }
 
-    @Override
-    public String toString() {
-        return "a="+Arrays.toString(matcher.values)+":ag="+Arrays.toString(aggregates)+":mt="+Arrays.toString(metrics);
-    }
-
```
fabuzaid21: why was this removed?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLMeanSummarizer.java
```diff
@@ -56,10 +61,10 @@
     public List<QualityMetric> getQualityMetricList() {
         List<QualityMetric> qualityMetricList = new ArrayList<>();
         qualityMetricList.add(
-                new SupportMetric(0)
+                new SupportQualityMetric(0)
```
fabuzaid21: This is more for @edgan8, but the way these QualityMetrics are instantiated seems confusing—what does the "0" mean here? (I know it's actually the index into aggregates, but I don't know why it has to be 0. The constructor isn't actually helpful here.) The QualityMetrics have a lot of implicit conventions that are just assumed to be followed, and there isn't much documentation for these conventions anywhere.

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLMeanSummarizer.java
```diff
@@ -29,6 +29,11 @@
         return Arrays.asList("count", "m1", "m2");
     }
 
+    @Override
+    public int[][] getEncoded(List<String[]> columns, DataFrame input) {
```
fabuzaid21: why does this take in `input` if it's not being used? If the abstract method specifies that it should be an argument, we should re-define the abstract method?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLOutlierSummarizer.java
```diff
@@ -24,6 +24,11 @@
         return Arrays.asList("Outliers", "Count");
     }
 
+    @Override
+    public int[][] getEncoded(List<String[]> columns, DataFrame input) {
```
fabuzaid21: Ah, so here you need `input`. Hmmmm………I don't have a good solution for this. I guess we can keep it as is

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLSummarizer.java
```diff
@@ -53,12 +52,10 @@ public void process(DataFrame input) throws Exception {
...
         encoder.setColumnNames(attributes);
         long startTime = System.currentTimeMillis();
-        List<int[]> encoded = encoder.encodeAttributes(
-                input.getStringColsByName(attributes)
-        );
+        int[][] encoded = getEncoded(input.getStringColsByName(attributes), input);
         long elapsed = System.currentTimeMillis() - startTime;
-        log.debug("Encoded in: {}", elapsed);
-        log.debug("Encoded Categories: {}", encoder.getNextKey());
+        log.info("Encoded in: {}", elapsed);
```
fabuzaid21: Nit: can we change this to "Encoded in: {} ms"—it's much more informative when people actually look at the logs

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APrioriLinear.java
```diff
@@ -19,20 +16,20 @@
  * are the most interesting as defined by "quality metrics" on these aggregates.
  */
 public class APrioriLinear {
-    Logger log = LoggerFactory.getLogger("APLSummarizer");
+    private Logger log = LoggerFactory.getLogger("APLSummarizer");
```
fabuzaid21: change to "APrioriLinear"

PR #245, opened 2018-02-04
Add cascades to APLinear

jialinding:


lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APrioriLinear.java
```diff
@@ -50,18 +52,50 @@ public APrioriLinear(
...
             final List<int[]> attributes,
             double[][] aggregateColumns
+    ) {
+        return explain(attributes, aggregateColumns, null);
+    }
+
+    public List<APLExplanationResult> explain(
+            final List<int[]> attributes,
+            double[][] aggregateColumns,
+            Map<String, int[]> aggregationOps
```
edgan8: Nice, I like this design of passing around aggregationOps. What do you think about passing around a AggregationOp[] array instead for an enum AggregationOp, I think it'll be faster and easier then you don't need three separate for loops you can just look up the appropriate operation for column j.

PR #248, opened 2018-02-07
Fixed a bug in CSV parsing exception handling

kraftp:


lib/src/main/java/edu/stanford/futuredata/macrobase/ingest/CSVDataFrameParser.java
```diff
@@ -93,10 +93,11 @@ public DataFrame load() throws Exception {
                         stringColumns[stringColNum++].add(rowValue);
                     } else if (t == Schema.ColType.DOUBLE) {
                         try {
-                            doubleColumns[doubleColNum++].add(Double.parseDouble(rowValue));
+                            doubleColumns[doubleColNum].add(Double.parseDouble(rowValue));
                         } catch (NumberFormatException e) {
-                            doubleColumns[doubleColNum++].add(Double.NaN);
+                            doubleColumns[doubleColNum].add(Double.NaN);
```
edgan8: Can you add a log message that displays the percent of rows that failed to parse when it's > 0? That way users still get visibility when the metrics are all the wrong type.

kraftp: Done!

PR #262, opened 2018-04-08
Bound max order in APL by number of attributes

jialinding:
The max combo order in APrioriLinear currently defaults to 3. This can cause lots of unnecessary computation when the actual number of attributes is fewer than the max combo order.

The current logic in APrioriLinear that deals with this scenario is weird: for example, if there are only two attributes (A1, A2) and we try to find order 3 results, APrioriLinear will attempt to find results composed of two values from A1 and one value from A2. Not sure why the code was written that way, but in any case this PR will avoid that scenario.


fabuzaid21:
Good catch, Jialin. This is because we changed the implementation of the pruning rules in the order-3 case: rather than iterate over the candidates in curOrderNext, we generate order-3 combinations and then call `validateCandidate` on it, to see if all order-2 subsets are contained.

@kraftp what was the justification for making this change in the pruning rule from order-2 to order-3. Not only does it cause bugs—it seems like it would fundamentally be slower.

PR #264, opened 2018-05-02
Mean shift movers query

kraftp:


lib/src/test/java/edu/stanford/futuredata/macrobase/analysis/summary/aplinear/APLCountMeanShiftSummarizerTest.java
```diff
@@ -0,0 +1,58 @@
...
+        APLCountMeanShiftSummarizer summ = new APLCountMeanShiftSummarizer();
+        summ.setMinSupport(.05);
+        summ.setMinMeanShift(1.1);
+        summ.setAttributes(explanationAttributes);
+        summ.process(output);
+        APLExplanation e = summ.getResults();
+        TestCase.assertEquals(3, e.getResults().size());
+        assertTrue(e.prettyPrint().contains("location=AUS"));
+        assertTrue(e.prettyPrint().contains("version=v2"));
+        assertTrue(e.prettyPrint().contains("language=ENG"));
```
edgan8: the fact that ENG shows up as an explanation attribute is worrisome: since every row is ENG the column should be irrelevant and should not be included in the result. Do you have a sense of what's going on?

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/util/qualitymetrics/MeanShiftQualityMetric.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public MeanShiftQualityMetric(
+            int oCountIdx,
+            int iCountIdx,
+            int oMeanIdx,
+            int iMeanIdx
+    ) {
+        this.oCountIdx = oCountIdx;
+        this.iCountIdx = iCountIdx;
+        this.oMeanIdx = oMeanIdx;
```
edgan8: I'm pretty sure this is also a sum / meanSum. Please make sure the naming is consistent otherwise it will get confusing what's actually a mean and what's actually a sum!

lib/src/main/java/edu/stanford/futuredata/macrobase/analysis/summary/util/qualitymetrics/MeanShiftQualityMetric.java
```diff
@@ -5,18 +5,18 @@
...
 public class MeanShiftQualityMetric implements QualityMetric {
-    private int oCountIdx, iCountIdx, oMeanIdx, iMeanIdx;
+    private int oCountIdx, iCountIdx, oMeanCountIdx, iMeanCountIdx;
 
     public MeanShiftQualityMetric(
             int oCountIdx,
             int iCountIdx,
-            int oMeanIdx,
-            int iMeanIdx
+            int oMeanCountIdx,
```
edgan8: omg why do you call it meanCount rather than meanSum or something

PR #272, opened 2019-02-19
Support for range query classifiers

edgan8:
Supports custom outlier classifiers based on ranges of strings or numeric quantities.

[[x], [a,b], [c,d]] matches for values v==x or a<=v<b or c<=v<d

See cube_range.json for example usage

### DIFF
--- a/core/demo/cube_range.json
+++ b/core/demo/cube_range.json
@@ -1,0 +1,20 @@
+{
+  "pipeline": "CubePipeline",
+  "inputURI": "csv://core/demo/sample_cubed.csv",
+  "classifier": "predicate",
+  "countColumn": "count",
+
+  "metric": "release",
+  "predicate": "str_range",
+  "cutoff": {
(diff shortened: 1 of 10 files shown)


edgan8:
@sahaana you might be most familiar with this code, can you take a look? Alternatively @kexinrong

core/demo/cube_range.json
```diff
@@ -0,0 +1,20 @@
+{
+  "pipeline": "CubePipeline",
+  "inputURI": "csv://core/demo/sample_cubed.csv",
+  "classifier": "predicate",
+  "countColumn": "count",
+
+  "metric": "release",
+  "predicate": "str_range",
+  "cutoff": {
+    "outlier": [["12-11-10","12-30-00"]]
```
sahaana: Also add a config with multiple?

edgan8: good point

